PR #35, opened 2022-09-21
Feature/OINK-726, Create an endpoint that returns true if the user is logged in and JWT token is valid and return an error when either of those conditions are not met.

pellerano:
- feat: added isUserAuthenticated endpoint.


juandatorr99:
Task linked: [OINK-726 isUserAuthenticated endpoint](https://app.clickup.com/t/24497898/OINK-726)

juandatorr99 (approved):

PR #86, opened 2023-05-18
set environment on deploy job

edwardtineob:
Set environment prop as production for deploy job on aws ec2 (production).


pellerano (approved):
LGTM!

PR #91, opened 2023-06-08
release funds endpoint done

edwardtineob:
This feature provide an endpoint to release funds when officer confirms the client DOP deposit is completed. These are some details:
- releaseFunds query added.
- releaseFunds function into transaction service added.
- transactionType Enums added.
- P2PTransferFundsI interface added.
- transfer funds function on wallet service updated to use alcancia DOP txn wallet as source wallet.


pellerano (approved):
LGTM!

PR #98, opened 2023-07-10
check if money sended is less than 0.01.

DHurtado714-itesm:
In case money sended is less than 0.01 USDC, notification is not done.


pellerano (approved):
LGTM!

PR #101, opened 2023-08-10
Feat/OINK-1210, Balance History endpoint

pellerano:
- feat: updated getUserBalanceHistory


edwardtineob:
Good

edwardtineob (approved):
Lgtm

src/metamap/metamap.service.ts
```diff
@@ -70,17 +70,17 @@ export class MetamapService {
     const [kycData, kycError] = await catchResult(this.retrieveResourceDataWithUid(uid));
     if (kycError || !kycData) throw new ApolloError(`${kycError}`);
     const { documents } = kycData;
-    const { country, subtype, fields } = documents[0];
+    const { country, subtype, type, fields } = documents[0];
     const documentNumber = fields.documentNumber.value;
-    let curp = subtype === 'cedula-de-identidad-y-electoral' ? null : fields.curp.value;
+    let curp = !(subtype === 'credencial-para-votar' || subtype === 'Credencial Para Votar') ? null : fields.curp.value;
```
xeno097, line 75: It would be nice if all the possible values of subtype were stored in constants to avoid having hardcoded strings.

xeno097 (changes requested):

src/metamap/metamap.service.ts
bujosa: Yeah like enum or something like that

xeno097 (approved):

bujosa (approved):

PR #102, opened 2023-08-24
Feat/OINK-1215, Migrate wallet keys from S3 to KMS

pellerano:


src/wallets/wallets.resolver.ts
```diff
@@ -27,4 +27,9 @@ export class WalletsResolver {
   transferFunds(@Context() ctx: ContextType, @Args('transferData') transferInput: TransferInput) {
     return this.walletsService.transferFunds({ srcUserId: ctx.req.user.userId, ...transferInput });
   }
+
+  // @Query(() => Boolean)
+  // migrateFromS3ToKMS() {
+  //   return this.walletsService.migrateFromS3ToKMS();
+  // }
```
xeno097, line 34: Should this be commented out?

pellerano:
Pueden ir aprobando si les parece bien, pero aun no le hagan merge. Aun quiero probar con transacciones para confirmar de que esta funcionando todo.

src/utils/constants/index.ts
```diff
@@ -2,3 +2,7 @@ export const METAMAP_SUBTYPES = {
   DO: ['cedula-de-identidad-y-electoral', 'Cédula De Identidad Y Electoral'],
   MX: ['credencial-para-votar', 'Credencial Para Votar'],
 };
+
+export const AWS_REGIONS = {
+  1: 'us-east-1',
+};
```
xeno097, line 8: Maybe this should be an environment variable

src/wallets/wallets.service.ts
```diff
@@ -239,7 +304,7 @@ export class WalletsService {
     };
 
     const quoteResponse = await this.zeroXProvider.getQuote(params);
-    const { priv } = await this.getKeysFromAWS(userId);
+    const { priv } = await this.getKeysFromKMS(userId);
```
xeno097, line 307: Maybe it would be a good idea to keep a call to the `getKeysFromAws` method in case the `getKeysFromKMS` does not find the private key and then remove it after we are sure that the migration was successful and there was no data loss.

src/wallets/wallets.service.ts
```diff
@@ -365,7 +430,7 @@ export class WalletsService {
 
     const { srcUserId, destUserId, token, amount } = data;
 
-    const src = deposit ? this.alcanciaDOPTxnPrivateKey : (await this.getKeysFromAWS(srcUserId)).priv;
+    const src = deposit ? this.alcanciaDOPTxnPrivateKey : (await this.getKeysFromKMS(srcUserId)).priv;
```
xeno097, line 433: Same here

src/wallets/wallets.service.ts
```diff
@@ -140,11 +141,74 @@ export class WalletsService {
...
+    if (missingValue.length > 0) {
+      const promisesArr = missingValue.map(id => this.getKeysFromAWS(id));
+      const values = await Promise.all(promisesArr);
+      const promisesArr2 = values.map(val => this.uploadKeysToKMS(val.priv, val.pub, val.id));
+      await Promise.all(promisesArr2);
+    }
+
+    return true;
+  }
+
```
xeno097, line 211: We should consider moving this logic to its own service when we start the refactor

xeno097 (changes requested):

src/wallets/wallets.service.ts
pellerano: I'll revert it back to the original function and rename it to S3 instead of AWS since that's more specific.

src/wallets/wallets.service.ts
pellerano: Same here.

src/wallets/wallets.service.ts
pellerano: I'll probably just move it to its own service file now.

src/alcancia/alcancia.service.ts
```diff
@@ -16,13 +16,15 @@ import { TransactionStatus } from 'src/utils/enums';
...
 import { Repository } from 'typeorm';
+import { AwsService } from 'src/aws/aws.service';
 
 @Injectable()
 export class AlcanciaService {
   constructor(
     @Inject(forwardRef(() => UsersService)) private usersService: UsersService,
     @Inject(forwardRef(() => TransactionsService)) private transactionsService: TransactionsService,
     @Inject(forwardRef(() => WalletsService)) private walletsService: WalletsService,
+    @Inject(forwardRef(() => AwsService)) private awsService: AwsService,
```
xeno097, line 27: You can simply inject the `AwsService` with `private awsService: AwsService` because the `AlcanciaService` does not have a cyclic dependency with the `AwsService`

src/wallets/wallets.module.ts
```diff
@@ -9,12 +9,14 @@ import { TypeOrmModule } from '@nestjs/typeorm';
...
 import { TransactionsModule } from 'src/transactions/transactions.module';
 import { Transaction } from 'src/transactions/entities';
+import { AwsModule } from 'src/aws/aws.module';
 
 @Module({
   imports: [
     forwardRef(() => UsersModule),
     forwardRef(() => NotificationsModule),
     forwardRef(() => TransactionsModule),
+    forwardRef(() => AwsModule),
```
xeno097, line 19: I think that you can remove the `forwardRef` call and simply put the `AswModule` in the imports array as there isn't a cyclic dependency between the `WalletModule` and the `AwsModule`

src/aws/aws.resolver.ts
```diff
@@ -0,0 +1,12 @@
+import { Query, Resolver } from '@nestjs/graphql';
+import { AwsService } from './aws.service';
+
+@Resolver()
+export class AwsResolver {
+  constructor(private readonly awsService: AwsService) {}
+
+  @Query(() => Boolean)
+  migrateFromS3ToKMS() {
+    if (process.env.NODE_ENV === 'DEV') return this.awsService.migrateFromS3ToKMS();
```
xeno097, line 10: You should have a default return statement as the query will throw an error because if the `NODE_ENV` var is not `DEV` then nothing will be returned by the method and the graphql server will complain. I suppose this is something that will be used to migrate the data from a S3 bucket to KMS and will only be run locally by someone after a production deployment but should we leave this query open to anyone to run it?

src/alcancia/alcancia.module.ts
```diff
@@ -6,12 +6,14 @@ import { TransactionsModule } from 'src/transactions/transactions.module';
...
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { Transaction } from 'src/transactions/entities';
+import { AwsModule } from 'src/aws/aws.module';
 
 @Module({
   imports: [
     forwardRef(() => UsersModule),
     forwardRef(() => TransactionsModule),
     forwardRef(() => WalletsModule),
+    forwardRef(() => AwsModule),
```
xeno097, line 16: I think that you can remove the `forwardRef` call and simply put the `AswModule` in the imports array as there isn't a cyclic dependency between the `AlcanciaModule` and the `AwsModule`

src/aws/aws.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    });
+  }
+
+  async getKeysFromKMS(userId?: string) {
+    const aliasListCmd = new ListAliasesCommand({});
+    const aliases = await kmsClient.send(aliasListCmd);
+
+    const { TargetKeyId: kmsKeyId } = aliases.Aliases.find(obj => obj.AliasName.includes(userId));
+
+    const kmsKeyDetails = new DescribeKeyCommand({ KeyId: kmsKeyId });
```
xeno097, line 32: Having a look at the docs for [ListAliasesCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-kms/classes/listaliasescommand.html) It seems that you can pass as filter the `KeyId` property so you can get the user's alias without having to filter the whole list after fetching all the records.

src/aws/aws.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      const aliasCmd = new CreateAliasCommand({
+        AliasName: `alias/${userId}`,
+        TargetKeyId: data.KeyMetadata.KeyId,
+      });
+
+      await kmsClient.send(aliasCmd);
+    });
+  }
+
+  async getKeysFromKMS(userId?: string) {
```
xeno097, line 26: can you add a type to describe the kind of output that this function will return? Something like:

```typescript
interface UserKeyPair {
    privateKey: string;
    publicKey: string;
}
```

src/aws/aws.service.ts
```diff
@@ -0,0 +1,118 @@
...
+import { ListObjectsCommand, PutObjectCommand, GetObjectCommand } from '@aws-sdk/client-s3';
+import { Injectable } from '@nestjs/common';
+import { SigningKey } from 'ethers';
+import { kmsClient, s3Client } from 'src/utils/libs/aws.clients';
+
+@Injectable()
+export class AwsService {
+  constructor() {}
+
+  async uploadKeysToKMS(privateKey: string | SigningKey, publicKey: string | SigningKey, userId: string) {
```
xeno097, line 11: Can you add typing here too?

src/aws/aws.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    const aliases = await kmsClient.send(aliasListCmd);
+
+    const { TargetKeyId: kmsKeyId } = aliases.Aliases.find(obj => obj.AliasName.includes(userId));
+
+    const kmsKeyDetails = new DescribeKeyCommand({ KeyId: kmsKeyId });
+
+    return await kmsClient.send(kmsKeyDetails).then(data => {
+      const keys = data.KeyMetadata.Description.split('/');
+      return { pub: keys[0], priv: keys[1] };
+    });
```
xeno097, line 37: Is there a reason why you are awaiting and chaining a `then` to the promise? I think you can simplify the statement by rewriting it like this:

```typescript
    const data = await kmsClient.send(kmsKeyDetails);

    const keys = data.KeyMetadata.Description.split('/');
    return { pub: keys[0], priv: keys[1] };
```

src/aws/aws.service.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    return await kmsClient.send(cmd).then(async data => {
+      const aliasCmd = new CreateAliasCommand({
+        AliasName: `alias/${userId}`,
+        TargetKeyId: data.KeyMetadata.KeyId,
+      });
+
+      await kmsClient.send(aliasCmd);
+    });
```
xeno097, line 23: Same `then` comment as below

xeno097 (changes requested):

src/aws/aws.service.ts
pellerano: The issue with that is that you need to know the `KeyId` value you want to look up, which we don't.

src/aws/aws.service.ts
pellerano: By typing, do you mean addiing a return type?

xeno097: yup

src/aws/aws.service.ts
xeno097: Can't we store the associated `targetKeyId` for a user on wallet creation? so we can avoid querying unnecessary data

pellerano: `targetKeyId` is auto-generated. If you want to store the associated `targetKeyId` on wallet creation we can either add another column to the user table in our DB or have an entirely separate table.

xeno097: Yeah, I'd prefer that because when the user base grows the amount of data fetched from was will also grow making the whole operation slower. I think that we can store it in the user table and then we start refactoring the backend we can create a new table that holds the wallet information for a user.

pellerano: Done.

xeno097: Nice will you be writing a migration to add that data to the new field or a script?

pellerano: Script most likely.

pellerano: Done. Added `walletKeyId` to the user entity.

xeno097 (approved):

PR #105, opened 2023-09-06
zendesk user jwt endpoint

edwardtineob:
* User controller added
* generate jwt for zendesk endpoint added


pellerano (approved):
LGTM!

src/users/users.service.ts
```diff
@@ -257,4 +259,15 @@ export class UsersService {
...
+  async generateSupportToken(id: string) : Promise<string>{
+    const user = await this.findOne(id);
+    const payload = {
+      user: `${user.name} ${user.surname}`,
+      email: user.email,
+      iat: (new Date().getTime() / 1000),
+      jti: uuid.v4()
+    }
+    return jwt.sign(payload, process.env.ZENDESK_SECRET);
+  }
```
xeno097, line 272: I think that you can use the `JwtService` from `@nestjs/jwt` which is already installed instead of using the `jsonwebtoken` package. You should inject the `JwtService` in the `UserService` class and then you can use it like this: `this.jwtService.sign(....)`. Also, consider reading env variables using the `ConfigService` instead of directly accessing them using `process.env...`

ref: https://github.com/nestjs/jwt

src/users/users.controller.ts
```diff
@@ -0,0 +1,23 @@
+import { Controller, HttpStatus, Post, Req, Res } from '@nestjs/common';
+import { Request, Response } from 'express';
+import { UsersService } from './users.service';
+
+@Controller('users')
+export class UsersController {
+    constructor(
+        private readonly usersService: UsersService
+    ) {
+        // THIS IS EMPTY
```
xeno097, line 10: I think that you can remove this comment here

src/users/users.controller.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    @Post()
+    async generateUserToken(@Req() request: Request, @Res() response: Response) {
+
+        if(typeof request.body.user_token != 'string') {
+            return response.status(HttpStatus.BAD_REQUEST).send({error: 'token must be string'});
```
xeno097, line 17: You should consider using `nestjs` built-in http errors instead of directly manipulating the response object because this conflicts with nestjs error and context handling logic.  If  you perform the change then you can simply return `{ jwt }` at the end of the method

ref: https://docs.nestjs.com/exception-filters#built-in-http-exceptions

xeno097 (changes requested):

src/users/users.service.ts
edwardtineob: I've replaced jsonwebtoken and now I'm using @nest/jwt. We don't still have ConfigService.

src/users/users.controller.ts
edwardtineob: Comment removed

src/users/users.controller.ts
edwardtineob: Now I'm using nest built-in http errors. Thank you!

src/users/users.controller.ts
```diff
@@ -0,0 +1,21 @@
...
+import { UsersService } from './users.service';
+
+@Controller('users')
+export class UsersController {
+    constructor(
+        private readonly usersService: UsersService
+    ) { }
+
+    @Post()
+    async generateUserToken(@Req() request: Request) {
```
xeno097, line 12: Why don't you use the `@Body()` decorator from nestjs instead? You can optionally supply a prop name to directly get the `user_token` property. Example:
```typescript
async generateUserToken(@Body("user_token") user_token?: string) {
 // Code here ...
}
```

ref: https://docs.nestjs.com/custom-decorators#param-decorators

xeno097 (changes requested):

xeno097 (approved):

PR #106, opened 2023-09-07
Redeem Referral Endpoint

pellerano:
- Adds the new redeemReferral endpoint for the ReferralModule
- Adds 2 new environment variables:
  - `REFERRAL_PAYOUT_AMOUNT`
  - `REFERRAL_PAYOUT_TOKEN`

- Adds the following new graphql operations:
    - redeemReward

NOTE 1: This PR adds 1 new environment variable that should be set in production before deploying the backend otherwise it will not start.


src/alcancia/alcancia.service.ts
```diff
@@ -194,4 +198,24 @@ export class AlcanciaService {
...
+        amount: REFERRAL_PAYOUT_INFO.amount
+      }, true)
+  
+      await this.viralLoopService.redeemReward({id: referralId})
+  
+      return true
+    } catch (error) {
+      return false
+    }
+  }
```
xeno097, line 220: Why don't you move this logic inside the `RedeemRewardUseCase` instead so that you don't have to introduce a circular dependency between the `AlcanciaModule` and the `ReferralModule`?

xeno097 (changes requested):

src/referral/use-cases/redeem-reward.use-case.ts
```diff
@@ -0,0 +1,38 @@
...
+        token: REFERRAL_PAYOUT_INFO.token,
+        amount: REFERRAL_PAYOUT_INFO.amount
+      }, true)
+  
+      await this.viralLoopService.redeemReward({id: referralId})
+  
+      return true
+    } catch (error) {
+      return false
+    }
```
xeno097, line 32: I think you can move the `payoutReward` method implementation in the `execute` method body  because the only thing that the `execute` method is doing is calling `payoutReward` which is kinda pointless

xeno097 (changes requested):

xeno097 (approved):

bujosa (approved):

PR #110, opened 2023-09-14
Fix/OINK-NA, Move Zendesk JWT logic to GQL Resolver.

pellerano:
- fix: moved zendesk logic from controller to resolver.
- chore: codebase restructure.


src/users/users.controller.ts
```diff
@@ -1,22 +1,8 @@
...
-    async generateUserToken(@Body('user_token') user_token?: string) {
-
-        if(typeof user_token != 'string') {
-            throw new BadRequestException('user_token must be string');
-        }
-        const token: string = user_token;
-        const jwt: string = await this.usersService.generateSupportToken(token);
-        return {jwt: jwt};
-    }
+  constructor(private readonly usersService: UsersService) {}
```
xeno097, line 7: Considering that this class is empty shouldn't we delete it?

juandatorr99 (approved):

xeno097 (approved):

xeno097 (commented):
Remember to remove the user controller from the user module

PR #113, opened 2023-09-22
Feat/OINK-1237, Gas Fee / Faucet Refactor

pellerano:


src/wallets/wallets.service.ts
```diff
@@ -202,40 +216,31 @@ export class WalletsService {
...
+        BigInt(aPolUSDC * 1000000),
         userId,
         process.env.USDC_CONTRACT_ADDRESS,
         process.env.AAVE_USDC_CONTRACT_ADDRESS
       );
-
-      USDCBalance = await this.getWalletBalancePolygon(walletAddress, process.env.USDC_CONTRACT_ADDRESS);
     }
 
-    const totalBalance = balanceEtherscan + cUSDBalance + mcUSDBalance + USDCBalance;
```
xeno097, line 229: Can you re-add this line. If the swap is performed the conversion rate between USDC and aPolUSDC is not exactly 1:1 so an incorrect balance will be reported to the user that is why I was refetching the USDC balance after the swap

src/wallets/wallets.service.ts
```diff
@@ -102,38 +102,52 @@ export class WalletsService {
...
+        address: process.env.MATIC_CONTRACT_ADDRESS,
+        abi: maticABI,
+        provider: this.maticProvider,
+        amount: parseEther('0.5'),
+        gasPrice: (await this.maticProvider.getFeeData()).gasPrice,
+        gasLimit: 25000,
+      },
+    };
+
+    return contractInfoMap[tokenLowerCase] || null;
```
xeno097, line 133: I think this is potentially dangerous as null could be returned where a null value wasn't expected we should either throw an error if the value is null or enforce with the types that only the keys defined in the `contractInfoMap` can be passed as function arguments. This can be achieved either with an enum or literal type

src/wallets/wallets.service.ts
```diff
@@ -65,34 +65,34 @@ export class WalletsService {
...
+      },
+      usdc: {
         address: process.env.USDC_CONTRACT_ADDRESS,
         decimals: 6,
         gasLimit: 200000,
         provider: this.maticProvider,
-      };
+      },
+    };
+    return contractAddressMapping[tokenLowerCase];
```
xeno097, line 95: same as the comment below

xeno097 (changes requested):

src/wallets/entities/wallet.interfaces.ts
```diff
@@ -1,6 +1,15 @@
...
 }
+
+export type Token = 'apolusdc' | 'usdc' | 'mcusd' | 'cusd';
+
+export enum TokenE {
+  aPolUSDC = 'apolusdc',
+  USDC = 'usdc',
+  mcUSD = 'mcusd',
+  cUSD = 'cusd',
+}
```
xeno097, line 15: This is a bit redundant we should either stick to the literal type or the enum. I'd prefer the enum. And to convert from a string to the corresponding value have a function that throws an error if the string given as argument does not correspond to any of the values specified in the enum:

```typescript
enum TokenEnum {
  aPolUSDC = 'apolusdc',
  USDC = 'usdc',
  mcUSD = 'mcusd',
  cUSD = 'cusd',
}

function convertStringToTokenEnum(input: string): TokenEnum {
  const lowerCaseInput = input.toLowerCase();

  switch (lowerCaseInput) {
    case TokenE.aPolUSDC:
    case TokenE.USDC:
    case TokenE.mcUSD:
    case TokenE.cUSD:
      return lowerCaseInput as TokenE;
    default:
      throw new Error(`Invalid token input: ${input}`);
  }
}
```

src/suarmi/suarmi.service.ts
```diff
@@ -154,13 +155,19 @@ export class SuarmiService {
 
       await this.transactionsService.createTransaction(transaction, uid);
       if (type === 'WITHDRAW') {
-        const needsGas = await this.walletService.tokenNeedsGas(from_currency, walletAddress);
-        if (needsGas) await this.walletService.useFaucet(from_currency, walletAddress);
+        const needsGas = await this.walletService.tokenNeedsGas(TokenE[from_currency], walletAddress);
+        if (needsGas) await this.walletService.useFaucet(TokenE[from_currency], walletAddress);
```
xeno097, line 159: When you implement the function remember to replace this notation because this can be undefined if `from_currency` is not one of the valid enum values

xeno097 (changes requested):

src/alcancia/alcancia.service.ts
```diff
@@ -99,14 +100,15 @@ export class AlcanciaService {
...
+        const token = convertStringToTokenEnum(from_currency);
+        const needsGas = await this.walletsService.tokenNeedsGas(TokenE[token], walletAddress);
+        if (needsGas) await this.walletsService.useFaucet(TokenE[token], walletAddress);
 
         const clabe = await this.usersService.getLastUsedBankAccount(userId);
         if (clabe !== bank_account) await this.usersService.updateLastUsedBankAccount(userId, bank_account);
 
         await this.walletsService.autoTransferFunds(
-          from_currency,
+          TokenE[token],
```
xeno097, line 111: Considering that you already performed the conversion calling the function, you can directly use the token variable

src/suarmi/suarmi.service.ts
```diff
@@ -154,13 +155,14 @@ export class SuarmiService {
...
-        if (needsGas) await this.walletService.useFaucet(from_currency, walletAddress);
+        const token = convertStringToTokenEnum(from_currency);
+        const needsGas = await this.walletService.tokenNeedsGas(TokenE[token], walletAddress);
+        if (needsGas) await this.walletService.useFaucet(TokenE[token], walletAddress);
 
         const clabe = await this.userService.getLastUsedBankAccount(uid);
         if (clabe !== bank_account) await this.userService.updateLastUsedBankAccount(uid, bank_account);
 
-        await this.walletService.autoTransferFunds(from_currency, walletAddress, data.address, uid, from_amount);
+        await this.walletService.autoTransferFunds(TokenE[token], walletAddress, data.address, uid, from_amount);
```
xeno097, line 165: Same here

src/transactions/transactions.service.ts
```diff
@@ -282,10 +283,11 @@ export class TransactionsService {
       const [transaction, err] = await catchResult(this.transactionRepository.findOneBy({ id: transactionId }));
       if (err) throw new ApolloError(`${err}`);
 
+      const token = convertStringToTokenEnum(transaction.targetAsset);
       const transfer = await this.walletsService.transferFunds(
         {
           destUserId: transaction.userID,
-          token: transaction.targetAsset,
+          token: TokenE[token],
```
xeno097, line 290: Same here

xeno097 (changes requested):

xeno097 (approved):

PR #116, opened 2023-10-10
Feat/OINK-1249 - Revert KMS

pellerano:
feat: removed KMS keys from codebase


xeno097 (changes requested):
I think the only thing that is missing is encrypting the private keys using a kms key before uploading them to s3, maybe we should create a new bucket migrate the encrypted private keys and configure the backend to point to that bucket and delete the old one and finally delete the keys from kms

src/aws/aws.service.ts
```diff
@@ -31,8 +38,30 @@ export class AwsService {
...
+        const decryptCmd = new DecryptCommand({
+          KeyId: KMS_ENCRYPTION_KEY,
+          CiphertextBlob: Buffer.from(privateKey, 'base64'),
+        });
+        const res = await kmsClient.send(decryptCmd);
+        return Buffer.from(res.Plaintext).toString();
+      default:
+        return 'Valid type was not passed in.';
+    }
+  }
```
xeno097, line 62: I'd honestly split this into 2 separate functions, one to decrypt and the other to encrypt the data to keep things nice and clear

src/schema.gql
```diff
@@ -231,6 +232,7 @@ type Query {
 
   """Returns missing user info fields."""
   updateUserFields: [String!]
+  uploadKeysToS3(id: String!, priv: String!, pub: String!): Boolean!
```
xeno097, line 235: I think this shouldn't be here I suppose

xeno097 (changes requested):

src/aws/aws.service.ts
```diff
@@ -37,30 +37,38 @@ export class AwsService {
...
+  //     case 'ENCRYPT':
+
+  //     case 'DECRYPT':
+
+  //     default:
+  //       return 'Valid type was not passed in.';
+  //   }
+  // }
+
+  // KMS Functions
```
xeno097, line 70: I think that this should have been deleted but survived ahaha

xeno097 (changes requested):

xeno097 (approved):

PR #117, opened 2023-10-18
Feat: Check if user is registered by email

DHurtado714-itesm:


src/users/api/graphql/users.resolver.ts
```diff
@@ -51,6 +51,12 @@ export class UsersResolver {
     return this.usersService.findOneByPhoneNumber(ctx.req.user.userId, phoneNumber);
   }
 
+  @Query(() => User, { description: 'Returns user based on email.' })
+  @UseGuards(JwtAuthGuard)
+  userByEmailSearch(@Context() ctx: ContextType, @Args('email') email: string) {
+    return this.usersService.findOneByEmail(email);
+  }
+
```
xeno097, line 59: Honestly, I'd prefer to only return a boolean value instead of the whole user object as what the front end needs is to know only if the user exists or not. The way this is implemented it might leak information that we don't want to expose to the public. 
I'd also rename the query to something like `userExists`

xeno097 (changes requested):

PR #120, opened 2023-11-04
feat/usdc.e to usdc migration

xeno097:
### Changelog

As the bridged USDC (now renamed [USDC.e](https://polygonscan.com/address/0x2791bca1f2de4661ed88a30c99a7a9449aa84174)) will be deprecated on November 10th, we need to migrate the user balances to the native [USDC](https://polygonscan.com/token/0x3c499c542cef5e3811e1192ce70d8cc03d5c3359). This PR implements the required changes to swap the users' balance to the new token.

- Adds swap logic using the 0x API to swap from USDC.e to USDC
- Refactors the `getWalletBalancePolygon` method to get a token balance directly from the blockchain 

NOTE: 
- This PR adds a new env variable called `NATIVE_USDC_CONTRACT_ADDRESS` its value is equal to the address of the native USDC token on polygon that you can see [here](https://polygonscan.com/token/0x3c499c542cef5e3811e1192ce70d8cc03d5c3359)


DHurtado714-itesm (approved):
Let's take into consideration the TODO and make a task to fix them

xeno097:
> Let's take into consideration the TODO and make a task to fix them

Yeah considering that I made that lambda for migrating the balances periodically we could use the same one to migrate balances directly from aave's USDC to native USDC and remove that logic from here

pellerano (approved):

PR #121, opened 2023-11-09
feat/cognito integration

xeno097:
### Changelog

This PR integrates the backend with cognito deprecating the old auth flow that used amplify. 

- Adds the `NewAuthResolver` to expose the new auth functions.
- Adds the `AwsCognitoAuthProvider` which is a wrapper around the `aws-sdk/client-cognito-identity-provider` library.
- Marks the old auth logic as deprecated.
- Adds a custom error to signal that the access token has expired. The error message in this scenario will be `TOKEN_EXPIRED_ERROR`. When this message is received by a client app then this app must provide the refresh token to the backend in order to receive a new access token

NOTE: 
- The old account deletion flow has been broken on purpose by throwing an error for the reasons mentioned in previous meeting


src/auth/api/graphql/auth.resolver.new.ts
```diff
@@ -0,0 +1,157 @@
...
+  public async completeForgotPasswordRequest(
+    @Args('input') input: CompleteForgotPasswordInput
+  ): Promise<AuthChallengeResponse> {
+    await this.authService.completeForgotPassword(input);
+
+    return {
+      status: true,
+    };
+  }
+
```
DHurtado714-itesm, line 142: Where we are going to send the user the verification code. SMS?

xeno097: It uses the Cognito pool configuration to send the code so it can be SMS or Email

DHurtado714-itesm (approved):

pellerano (approved):

PR #122, opened 2023-11-09
feat(secrets): AWS SDK and use ConfigService for port and AWS

DHurtado714-itesm:
## Title:
Implement AWS SDK for managing `.env` keys using secrets manager

## Description:
This PR introduces a new way of storing important keys in the cloud. For now this is the base implementation, later on changes are going to be made in each module.


src/config/secrets.config.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default async () => {
+  const secretName = process.env.SECRETS_GROUP;
+  const data = await client.getSecretValue({ SecretId: secretName }).promise();
+  if (data.SecretString) {
+    return JSON.parse(data.SecretString);
+  } else {
+    throw new ApolloError('Secrets could not be retrieved');
+  }
+};
```
xeno097, line 14: I think that it would be a good idea to create a ts type to represent the configuration object, also, as the client is only used to connect to Secrets Manager I'd create it inside the function body.

src/main.ts
```diff
@@ -1,17 +1,17 @@
...
 async function bootstrap() {
-  const port = process.env.PORT;
-  const host = process.env.HOST || 'localhost';
   const app = await NestFactory.create(AppModule);
-
+  const configService = app.get(ConfigService);
+  const host = process.env.HOST || 'localhost';
+  const port = configService.get<string>('PORT');
+  const awsconfig = JSON.parse(configService.get<string>('AWS_CONFIG'));
   Amplify.configure(awsconfig);
```
xeno097, line 15: There is a file in the `config` folder called `constants` with an enum called `EnvConfigKey` with various keys used to load the env variables, we can start adding them to the enum to avoid hardcoded strings.

image

src/app.module.ts
```diff
@@ -9,6 +9,7 @@ import { AlcanciaModule } from './alcancia/alcancia.module';
 import { APYModule } from './apy/apy.module';
 import { AuthModule } from './auth/auth.module';
 import { AwsModule } from './aws/aws.module';
+import secretsConfig from './config/secrets.config';
```
xeno097, line 12: Just to be a bit more specific I'd call this `awsSecretConfigLoader`

xeno097 (changes requested):

src/config/IConfigType.config.ts
```diff
@@ -0,0 +1,95 @@
...
+  moralis: {
+    apiKey: string;
+    streamId: string;
+  };
+  ox: {
+    apiKey: string;
+    viralLoopApiKey: string;
+    viralLoopBaseUri: string;
+  };
+};
```
xeno097, line 95: I'd rename the type to `AppConfig`, also please update the file name to `app-config.interface.ts`

xeno097 (changes requested):

xeno097 (approved):

src/notifications/notifications.service.ts
```diff
@@ -4,21 +4,24 @@ import { ConfigService } from '@nestjs/config';
...
   @Inject(forwardRef(() => WalletsService)) private walletsService: WalletsService;
   @Inject(forwardRef(() => UsersService)) private usersService: UsersService;
   private httpService = new HttpService();
   private endpoint: string;
   private headers: { headers: { Authorization: string; 'Content-Type': string } };
 
   constructor(private configService: ConfigService) {
-    this.endpoint = this.configService.get<string>('FIREBASE_CLOUD_MESSAGING_URL');
-    const firebaseKey = this.configService.get<string>('FIREBASE_CLOUD_MESSAGING_KEY');
+    this.firebase = this.configService.get<FirebaseConfig>('firebase');
```
xeno097, line 22: I think you forgot to replace the hardcoded string with the equivalent `EnvConfigKey` value

src/main.ts
```diff
@@ -10,7 +10,7 @@ async function bootstrap() {
   const app = await NestFactory.create(AppModule);
   const configService = app.get(ConfigService);
   const host = process.env.HOST || 'localhost';
-  const port = configService.get<string>('PORT');
+  const port = configService.get<string>('port');
```
xeno097, line 13: I think you forgot to replace the hardcoded string with the equivalent `EnvConfigKey` value

src/utils/constants/constants.module.ts
```diff
@@ -0,0 +1,23 @@
...
+        {
+          provide: Constants,
+          useFactory: (configService: ConfigService): Constants => {
+            return new Constants(configService);
+          },
+          inject: [ConfigService],
+        },
+      ],
+      exports: [Constants],
+    };
```
xeno097, line 21: If you set this module as global you won't need to add the `Constants` provider to every `providers` array in every module where it is needed

xeno097 (changes requested):

xeno097 (approved):

PR #132, opened 2023-12-12
OINK-1329 - Transaction Module Refactor

pellerano:


src/modules/transaction/api/graphql/transactions.resolver.new.ts
```diff
@@ -0,0 +1,17 @@
...
+
+@Resolver(() => Transaction)
+export class TransactionsResolver {
+  constructor(private readonly withdrawUseCase: WithdrawUseCase) {}
+
+  @Mutation(() => Order)
+  public async withdraw(@Args('input') input: OrderInput, @Context() ctx: ContextType): Promise<Order> {
+    const { userId } = ctx.req.user;
+    return this.withdrawUseCase.execute({ input, userId });
+  }
```
xeno097, line 16: There is a new decorator called `JwtPayload` that returns the user information, please use that instead of the `Context` decorator. Also is this a crypto withdrawal or a bank withdrawal? The 2 withdrawal flows should be separated and we should try to be as explicit as possible. Considering the context of the PR you are working on the bank withdrawals so this should be called something like `withdrawToBankAccount`.

src/modules/transaction/core/use-cases/withdraw.use-case.ts
```diff
@@ -0,0 +1,21 @@
...
+  constructor(
+    private readonly drWithdrawUseCase: DRWithdrawUseCase,
+    private readonly mxWithdrawUseCase: MXWithdrawUseCase
+  ) {}
+
+  public async execute(input: WithdrawUseCaseDto) {
+    const { from_currency, to_currency } = input.input;
+    if (from_currency === 'MXN' || to_currency === 'MXN') return await this.mxWithdrawUseCase.execute(input);
+    if (from_currency === 'DOP' || to_currency === 'DOP') return await this.drWithdrawUseCase.execute(input);
+  }
```
xeno097, line 20: This should throw an error if no valid currency is provided. Instead of having hardcoded strings, we should have a `SupportedCurrency` enum to describe all the possible values for a currency that we currently support.

```typescript
export enum SupportedCurrency {
    MXN = "MXN",
    DOP = "DOP
}
```

Don't forget to add types everywhere as looking at the code as it is, it is not clear what the return type of this use case is and we also lose type safety

src/modules/transaction/core/use-cases/withdraw.use-case.ts
```diff
@@ -0,0 +1,21 @@
+import { OrderInput } from '../../api/graphql/input-types/order-input';
+import { DRWithdrawUseCase } from './dr-withdraw.use-case';
+import { MXWithdrawUseCase } from './mx-withdraw.use-case';
+
+export interface WithdrawUseCaseDto {
+  input: OrderInput;
+  userId: string;
+}
```
xeno097, line 8: Consider creating a type that has unnested fields. It should have all the fields in the OrderInput type + the userId field. Remember that we should not use types from the API layer in the business logic layer.

src/modules/transaction/core/use-cases/mx-withdraw.use-case.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Transaction } from '../../infra/database/postgres/models/transaction.entity';
+import { NewSuarmiService } from '../services/suarmi.service.new';
+import { NewTransactionService } from '../services/transaction.service.new';
+import { WithdrawUseCaseDto } from './withdraw.use-case';
+
+export class MXWithdrawUseCase {
+  constructor(
+    private readonly suarmiService: NewSuarmiService,
+    private readonly txnService: NewTransactionService,
+    @InjectRepository(Transaction) private txnRepository: Repository<Transaction>
```
xeno097, line 12: Please remember to implement an abstraction over typeorm own repository class instead of using it directly.

src/modules/transaction/core/use-cases/dr-withdraw.use-case.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Transaction } from '../../infra/database/postgres/models/transaction.entity';
+import { NewAlcanciaService } from '../services/alcancia.service.new';
+import { NewTransactionService } from '../services/transaction.service.new';
+import { WithdrawUseCaseDto } from './withdraw.use-case';
+
+export class DRWithdrawUseCase {
+  constructor(
+    private readonly alcanciaService: NewAlcanciaService,
+    private readonly txnService: NewTransactionService,
+    @InjectRepository(Transaction) private txnRepository: Repository<Transaction>
```
xeno097, line 12: Same comment as below, also try to avoid abbreviations when naming variables to make things as clear as possible. Example `txnRepository` should be `transactionRepository`

src/modules/transaction/core/services/transaction.service.new.ts
```diff
@@ -0,0 +1,36 @@
...
+      TransactionStatus['PENDING'],
+      null,
+      null,
+      null,
+      null,
+      null,
+      method
+    );
+  }
+}
```
xeno097, line 36: This should be a repository as you are using database-specific objects, try to use objects instead of a long list of parameters. When using enums try to not access them using the index syntax. `TransactionStatus["PENDING"]` should be  `TransactionStatus.PENDING`.

src/modules/transaction/core/services/alcancia.service.new.ts
```diff
@@ -0,0 +1,62 @@
...
+      formattedOrderInput;
+    if (type === 'WITHDRAW' && !bank_account)
+      throw new ApolloError('Por favor enviar una cuenta bancaria para los retiros.');
+    const concepto = uuid().split('-')[4].substring(0, 7);
+
+    const method = {};
+
+    if (!!bank_name) method['bankName'] = bank_name;
+    if (!!bank_account) method['bankAccount'] = bank_account;
+    method['concepto'] = concepto;
```
xeno097, line 35: Create a type for the method variable and avoid using indexed access.

src/modules/transaction/core/services/alcancia.service.new.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const txnInput: TransactionInput = {
+      amount: parseFloat(to_amount),
+      sourceAmount: parseFloat(from_amount),
+      type: type,
+      provider: 'ALCANCIA',
```
xeno097, line 41: We should add an enum to describe all the kind of provider that we support instead of using hardcoded string

src/modules/transaction/core/services/alcancia.service.new.ts
```diff
@@ -0,0 +1,62 @@
...
+      response.to_currency = 'usdc';
+    return response;
+  }
+
+  public async sendOrder(orderInput: OrderInput) {
+    const formattedOrderInput = await this.formatOrderInput(orderInput);
+
+    const { type, bank_name, bank_account, from_amount, from_currency, to_currency, to_amount, network } =
+      formattedOrderInput;
+    if (type === 'WITHDRAW' && !bank_account)
```
xeno097, line 27: We should create an enum for the supported types of transactions

src/modules/transaction/core/services/alcancia.service.new.ts
```diff
@@ -0,0 +1,62 @@
...
+    const response = { ...orderInput };
+    if (this.constants.DEPRECATED_CURRENCIES.includes(orderInput.from_currency.toLowerCase()))
+      response.from_currency = 'usdc';
+    if (this.constants.DEPRECATED_CURRENCIES.includes(orderInput.to_currency.toLowerCase()))
+      response.to_currency = 'usdc';
+    return response;
+  }
+
+  public async sendOrder(orderInput: OrderInput) {
+    const formattedOrderInput = await this.formatOrderInput(orderInput);
```
xeno097, line 23: Remember to add types to function signatures

src/modules/transaction/core/services/suarmi.service.new.ts
```diff
@@ -0,0 +1,168 @@
...
+      type: type,
+      provider: 'SUARMI',
+      paymentGatewayID: data.uuid,
+      targetAsset: to_currency,
+      sourceAsset: data.from_currency,
+      secret: '',
+      method: JSON.stringify({ ...method }),
+    };
+    return { txnInput: { ...txnInput }, data };
+  }
```
xeno097, line 167: Please remember to only put logic that communicates with the suarmi API here. As suarmi is an external depency it should be moved to the infra layer and be "downgraded" to a provider instead of being a service

xeno097 (changes requested):
You are on the right track but remember to use the base errors like `BaseInternalError` and `BaseError` to create errors instead of the Apollo error class as that is something specific to a graphql API and should not be used in the core layer.

src/modules/transaction/core/services/alcancia.service.new.ts
pellerano: Used existing AllowedProviders enum and added ALCANCIA as an option.

src/modules/transaction/core/services/suarmi.service.new.ts
```diff
@@ -0,0 +1,171 @@
...
+      state: state,
+      zip,
+      curp: curp,
+      'id-number': idnumber,
+    };
+    return userData;
+  }
+
+  private async suarmiUserExists(user: User): Promise<string> {
+    try {
```
DHurtado714-itesm, line 97: Maybe you should check if the user is mexican, before splitting the phone number.

src/modules/transaction/core/services/suarmi.service.new.ts
```diff
@@ -0,0 +1,171 @@
...
+      const suarmiUserId = await this.suarmiUserExists(user);
+      if (suarmiUserId) return suarmiUserId;
+
+      const data = await this.formatSuarmiUserCreationData(user);
+
+      const url = `${this.suarmi.apiUrl}/user`;
+      const { uuid } = (await lastValueFrom(this.httpService.post(url, data, this.getHeaders()))).data;
+      await this.userRepository.save({ ...user, suarmi: uuid });
+      return uuid;
+    } catch (e: any) {
```
DHurtado714-itesm, line 118: Keep consistency in error variable, choose error or e, not both.

src/modules/transaction/core/services/suarmi.service.new.ts
```diff
@@ -0,0 +1,171 @@
...
+          {
+            ...suarmiOrderInput,
+            user: suarmi,
+            to_address: type === SupportedTransactionType.DEPOSIT ? walletAddress : bank_account,
+          },
+          this.getHeaders()
+        )
+        .pipe()
+    );
+    if (res.status != 201) throw new ApolloError('Error al crear orden de compra en Suarmi');
```
DHurtado714-itesm, line 148: As suggestion, maybe we could also throw the error code for better debugging

src/modules/transaction/core/services/transactions.service.ts
```diff
@@ -276,9 +276,9 @@ export class TransactionsService {
   async formatOrderInput(orderInput: OrderInput): Promise<OrderInput> {
     const response = { ...orderInput };
     if (this.constants.DEPRECATED_CURRENCIES.includes(orderInput.from_currency.toLowerCase()))
-      response.from_currency = 'usdc';
+      response.from_currency = SupportedCurrency.USDC;
```
DHurtado714-itesm, line 279: I like this 😍

src/modules/transaction/core/use-cases/dr-withdraw.use-case.ts
```diff
@@ -0,0 +1,26 @@
...
+import { NewTransactionService } from '../services/transaction.service.new';
+import { WithdrawUseCaseDto } from './withdraw.use-case';
+
+export class DRWithdrawUseCase {
+  constructor(
+    private readonly alcanciaService: NewAlcanciaService,
+    private readonly txnService: NewTransactionService,
+    @InjectRepository(TransactionDBModel) private transactionRepository: Repository<TransactionDBModel>
+  ) {}
+
```
DHurtado714-itesm, line 14: If you are going to comment the code, do it in a specific format. We should discuss it with @xeno097

DHurtado714-itesm (changes requested):

src/modules/transaction/core/services/suarmi.service.new.ts
pellerano: If its suarmi, then it's mexico.

DHurtado714-itesm (approved):

src/modules/transaction/core/domain/entities/transaction.entity.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * The amount the user is providing for the transaction.
+   */
+  sourceAmount?: number;
+
+  /**
+   * The identifier of the user associated with this transaction.
+   */
+  userID?: EntityId;
```
xeno097, line 77: please rename this to `userId`

src/modules/transaction/core/domain/entities/transaction.entity.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * The ending currency for the transaction.
+   */
+  targetAsset?: string;
+
+  /**
+   * Deprecated: entity id used when Firebase was in place.
+   */
+  transactionID?: EntityId;
```
xeno097, line 42: Please rename this to `transactionId`

src/modules/transaction/core/domain/entities/transaction.entity.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Date at which transaction was created.
+   */
+  createdAt: Date;
+
+  /**
+   * The external identifier for third-party service providers.
+   */
+  paymentGatewayID?: EntityId;
```
xeno097, line 22: Please rename this to `paymentGatewayId`

src/modules/transaction/infra/libs/suarmi/suarmi.provider.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  public async sendOrder(input: WithdrawUseCaseDto): Promise<{ entity: TransactionEntity; data: Order }> {
+    const { type, bank_account, from_amount, to_amount, bank_name, userId: uid } = input;
+    const { to_currency } = await this.formatOrderInput(input);
+
+    if (type === TransactionType.WITHDRAW && !bank_account) throw new InvalidBankAccountError();
+
+    const user = await this.userRepository.findOneBy({ id: uid });
+    let { suarmi } = user;
+    const { walletAddress } = user;
```
xeno097, line 134: you should pass these values to the `sendOrder` method instead of having to fetch them here

src/modules/transaction/infra/libs/suarmi/suarmi.provider.ts
```diff
@@ -0,0 +1,177 @@
...
+    } catch (error: any) {
+      throw new SuarmiUserCreationError();
+    }
+  }
+
+  public async sendOrder(input: WithdrawUseCaseDto): Promise<{ entity: TransactionEntity; data: Order }> {
+    const { type, bank_account, from_amount, to_amount, bank_name, userId: uid } = input;
+    const { to_currency } = await this.formatOrderInput(input);
+
+    if (type === TransactionType.WITHDRAW && !bank_account) throw new InvalidBankAccountError();
```
xeno097, line 130: you should create a type that does not allow the bank account value to be undefined considering that this is required for performing a bank withdrawal. Just to clarify something, with suarmi can you create other types of orders or can you only withdraw to a bank?

src/modules/transaction/core/use-cases/withdraw.use-case.ts
```diff
@@ -0,0 +1,32 @@
...
+      throw new InvalidCurrencyError();
+    if (!Object.values(SupportedFiatCurrency).includes(to_currency as SupportedFiatCurrency))
+      throw new InvalidCurrencyError();
+
+    if (from_currency === SupportedFiatCurrency.MXN || to_currency === SupportedFiatCurrency.MXN)
+      return await this.mxWithdrawUseCase.execute(input);
+
+    if (from_currency === SupportedFiatCurrency.DOP || to_currency === SupportedFiatCurrency.DOP)
+      return await this.drWithdrawUseCase.execute(input);
+  }
```
xeno097, line 31: You should throw an error if the end of the method is reached because this means that none of the previous cases has been matched

src/modules/transaction/infra/libs/suarmi/suarmi.provider.ts
```diff
@@ -0,0 +1,177 @@
...
+      colonia,
+      'ext-num': extNum,
+      'int-num': intNum,
+      state: state,
+      zip,
+      curp: curp,
+      'id-number': idnumber,
+    };
+    return userData;
+  }
```
xeno097, line 97: this should be an external function instead of a class method. Also, define a return type for this function.

src/modules/transaction/infra/libs/suarmi/suarmi.provider.ts
```diff
@@ -0,0 +1,177 @@
...
+  private async createUser(user: UserDBModel): Promise<string> {
+    try {
+      const suarmiUserId = await this.suarmiUserExists(user);
+      if (suarmiUserId) return suarmiUserId;
+
+      const data = await this.formatSuarmiUserCreationData(user);
+
+      const url = `${this.suarmi.apiUrl}/user`;
+      const { uuid } = (await lastValueFrom(this.httpService.post(url, data, this.getHeaders()))).data;
+      await this.userRepository.save({ ...user, suarmi: uuid });
```
xeno097, line 119: You shouldn't be using the `userRepository` here. Saving the user data after the user has been created in the 3rd party API should be done in the use case/ service method that requires the user to be created in the 3rd party service

src/modules/transaction/infra/libs/suarmi/suarmi.provider.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    const method = {
+      concepto: data.concepto,
+    };
+
+    if (!!bank_name) method['bankName'] = bank_name;
+    if (!!bank_account) method['bankAccount'] = bank_account;
+
+    const entity: TransactionEntity = {
+      id: EntityId.fromString(uuid()),
```
xeno097, line 162: you should be using the `EntityId.new` method if you want a new uuid.

xeno097 (changes requested):

src/modules/transaction/api/graphql/input-types/method.input.ts
```diff
@@ -2,9 +2,12 @@ import { Field, InputType } from '@nestjs/graphql';
...
+  @Field({ nullable: true })
+  bankName?: string;
+
+  @Field({ nullable: true })
+  bankAccount?: string;
 
   @Field()
-  bankAccount: string;
+  concepto: string;
 }
```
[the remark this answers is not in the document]
pellerano: I think this might cause some issues, since the value is being used in the DB.

src/modules/transaction/core/services/transactions.service.ts
```diff
@@ -276,9 +276,9 @@ export class TransactionsService {
   async formatOrderInput(orderInput: OrderInput): Promise<OrderInput> {
     const response = { ...orderInput };
     if (this.constants.DEPRECATED_CURRENCIES.includes(orderInput.from_currency.toLowerCase()))
-      response.from_currency = 'usdc';
+      response.from_currency = SupportedCryptoCurrency.USDC;
     if (this.constants.DEPRECATED_CURRENCIES.includes(orderInput.to_currency.toLowerCase()))
-      response.to_currency = 'usdc';
+      response.to_currency = SupportedCryptoCurrency.USDC;
     return response;
```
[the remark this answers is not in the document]
pellerano: changed back to 'usdc'

src/modules/transaction/infra/libs/suarmi/suarmi.provider.ts
pellerano: both withdrawal and deposits for MX are managed through Suarmi.

src/modules/transaction/api/graphql/input-types/method.input.ts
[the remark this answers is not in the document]
xeno097: The point of splitting types per layer is exactly to avoid that you can have that field be called `concepto` in the database but that can have a completely different name in the other layers

src/modules/transaction/infra/libs/suarmi/suarmi.provider.ts
xeno097: yup you should do that please because it's clearer that way and we can ensure that each flow has the expected type

src/modules/transaction/core/infra/suarmi-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+  from_currency: SupportedFiatCurrency | SupportedCryptoCurrency;
+  from_amount: string;
+  to_currency: SupportedFiatCurrency | SupportedCryptoCurrency;
+  to_amount: string;
+  network: string;
+  type: string;
+  bank_account?: string;
+  bank_name?: string;
+  user: UserEntity;
+}
```
xeno097, line 16: Based on my comment in the suarmi provider file this should be split into 2 types one for withdrawals and one for deposits

src/modules/transaction/core/infra/suarmi-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+  phone_code: string;
+  phone_number: string;
+  street: string;
+  colonia: string;
+  'ext-num': string;
+  'int-num': string;
+  state: string;
+  zip: string;
+  curp: string;
+  'id-number': string;
```
xeno097, line 41: you should rename the fields like `ext-num`  and the others that are in `snake_case` to be in `camelCase` and if that is the format expected or received by the suarmi API then implement an intermediary type that has the required format and parse it. As an example have a look at the 0x integration and the distinction between raw types from the API and our internal types

xeno097 (changes requested):

src/modules/transaction/infra/libs/suarmi/entities/suarmi-user.entity.ts
```diff
@@ -0,0 +1,13 @@
...
+  /**
+   *  Banking code for mexican users
+   */
+  curp: string;
+
+  /**
+   * User document number associated for their kyc.
+   */
+  documentNumber: string;
+}
```
xeno097, line 13: This is not allowed as you are importing an interface from another module

src/modules/transaction/infra/repositories/transaction.repository.ts
```diff
@@ -108,7 +103,28 @@ export class TransactionRepository extends ITransactionRepository {
...
+          conversionRate: entity.conversionRate,
+          currentBalance: entity.currentBalance,
+          newBalance: entity.newBalance,
+          sourceAmount: entity.sourceAmount,
+          provider: entity.provider,
+          method: entity.method,
+          receiverId: `${entity.receiverId}`,
+          senderId: `${entity.senderId}`,
+          txnHash: entity.txnHash,
+        }
```
xeno097, line 126: Should all these fields be updated? I don't think that we should change the `type` of a transaction once it has been created nor the `createdAt` field

xeno097 (changes requested):

xeno097 (approved):

src/modules/transaction/core/use-cases/mx-withdraw.use-case.ts
```diff
@@ -0,0 +1,73 @@
...
+import { WithdrawUseCaseDto } from './withdraw.use-case';
+
+export class MXWithdrawUseCase {
+  constructor(
+    private readonly suarmiProvider: SuarmiProvider,
+    private constants: Constants,
+    @InjectRepository(TransactionDBModel) private transactionRepository: ITransactionRepository,
+    @InjectRepository(UserDBModel) private userRepository: IUserRepository,
+    @InjectRepository(KycDBModel) private kycRepository: IKycRepository
+  ) {}
```
xeno097, line 25: Same comment about inject repository as the others

src/modules/transaction/infra/libs/alcancia/alcancia.provider.ts
```diff
@@ -0,0 +1,42 @@
...
+        concept,
+        from_amount,
+        from_currency,
+        network,
+        to_currency,
+        to_amount,
+        uuid: uuid(),
+      },
+    };
+  }
```
xeno097, line 41: Return types are missing in both methods

src/modules/transaction/core/use-cases/get-balance-history.use-case.ts
```diff
@@ -0,0 +1,49 @@
...
+import { BalanceHistoryDBModel } from '../../infra/database/postgres/models/balance-history.model';
+import { IBalanceHistoryRepository } from '../../infra/repositories/balance-history.repository';
+import { BalanceHistoryEntity } from '../domain/entities/balance-history.entity';
+
+interface GetBalanceHistoryDto {
+  userId: string;
+}
+
+export class GetBalanceHistoryUseCase {
+  constructor(@InjectRepository(BalanceHistoryDBModel) private balanceHistoryRepository: IBalanceHistoryRepository) {}
```
xeno097, line 12: Same comment about inject repository

src/modules/transaction/core/use-cases/get-balance-history.use-case.ts
```diff
@@ -0,0 +1,49 @@
+import { InjectRepository } from '@nestjs/typeorm';
+import { EntityId } from 'src/core/domain/entity/entity-id';
+import { BalanceHistoryDBModel } from '../../infra/database/postgres/models/balance-history.model';
+import { IBalanceHistoryRepository } from '../../infra/repositories/balance-history.repository';
+import { BalanceHistoryEntity } from '../domain/entities/balance-history.entity';
+
+interface GetBalanceHistoryDto {
+  userId: string;
+}
```
xeno097, line 9: Please rename the `userId` field to `id` and the type to `GetBalanceHsitoryByUserIdDto`

xeno097 (changes requested):

.vscode/settings.json
```diff
@@ -8,5 +8,13 @@
...
+  "workbench.colorCustomizations": {
+    "window.inactiveBorder": "#cccccc",
+    "window.activeBorder": "#295c8a",
+    "activityBar.background": "#1e2d55b7",
+    "titleBar.activeBackground": "#4e76e58a",
+    "titleBar.activeForeground": "#F4FCF3"
+  },
+  "conventionalCommits.scopes": ["suarmi", "metamap"],
+  "typescript.tsdk": "node_modules\\typescript\\lib"
 }
```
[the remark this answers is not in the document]
pellerano: It's already in the `.gitignore`

src/modules/transaction/infra/libs/alcancia/alcancia.provider.ts
pellerano: I took a closer look at the code, and this is not being used anywhere 🤔, I'll still add the return types to their appropriate use cases.

PR #134, opened 2023-12-17
feat/user module refactor part 2

xeno097:
### Changelog

This PR focuses on the refactoring the user module `core` and `api` layers

- Adds input validations for the auth resolver inputs
- Distinguishes between graphql types and other classes
- Implements mapper utility functions to map from entities to dto

Notes: 
- As the mobile app still uses legacy code, what is still missing to be refactored will be updated after the app migrates to use te cognito auth flow


pellerano (approved):

src/modules/user/core/dto/user.dto.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * The user birthdate.
+   */
+  birthdate?: Date;
+
+  /**
+   * The user telephone number.
+   */
+  phoneNumber?: string;
```
DHurtado714-itesm, line 60: Maybe this should be mandatory

DHurtado714-itesm (approved):

src/modules/user/core/dto/user.dto.ts
xeno097: it should but after verifying in the database i saw that there users without phone number. When we start working on the database refactor we should address this

PR #135, opened 2023-12-19
Refactor Metamap to Modular Monolith Architecture

DHurtado714-itesm:
Description:

Refactor KYC provider logic into 3 layers(Repository, Service, Metamap provider). 

News:
- `resourceId` attribute added to staging DB
- `request` table deprecated

To keep in mind: A migrations has to be done, before deploying to prod


src/modules/user/core/services/kyc.service.new.ts
pellerano, line 1: This file feels redundant since @xeno097 already created a `kyc.repository.ts` that essentially works the same 🤔

src/utils/constants/index.ts
```diff
@@ -14,6 +14,11 @@ export class Constants {
     MX: ['credencial-para-votar', 'Credencial Para Votar'],
   };
 
+  METAMAP_STATUS = {
+    REVIEW_NEEDED: 'pending',
```
pellerano, line 18: Shouldn't this be `PENDING`?

src/modules/user/api/webhook/metamap.webhook.ts
```diff
@@ -1,13 +1,12 @@
...
 
 @Controller('metamap')
 export class MetamapController {
-  constructor(private metamapService: MetamapService) {}
+  constructor(private kycService: KycService) {}
 
   @Post('/verification/complete')
-  completeVerification(@Req() request: Request) {
-    return this.metamapService.retrieveResourceData(request);
+  async completeVerification(@Req() request: Request) {
```
xeno097, line 9: can you create a type definition for the expected payload received by metamap when calling this endpoint? Using the `Request` type from express/nestjs does not help us with type safety

src/modules/user/core/errors/Metamap/internal-server.error.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseError } from 'src/core/errors/base.error';
+
+export class InternalServerError extends BaseError {
+  constructor() {
+    super({
+      code: 'INTERNAL_SERVER_ERROR',
+      message: 'An internal server error has occurred.',
+    });
+  }
+}
```
xeno097, line 10: This can be removed has the `core` folder already defines an internal server error
image

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,183 @@
...
+    const documentNumber = fields.documentNumber.value;
+    const curp = !this.constants.METAMAP_SUBTYPES.MX.includes(subtype) ? fields.curp.value : null;
+    let address = null;
+
+    if (this.constants.METAMAP_SUBTYPES.MX.includes(subtype)) {
+      address = fields.address.value;
+    }
+
+    const userIdEntity = EntityId.fromString(uid);
+    const currentKyc = await this.kycRepository.getByUserId(userIdEntity);
```
xeno097, line 47: Consider that this will throw an error if the kyc data does not exist for the provided user id. Considering that this method is called createKycData this should verify if kyc exists with the provided user id and throw an error if it does because is not the expected behaviour when you try to create a resource.
image

src/modules/user/infra/repositories/kyc.repository.ts
```diff
@@ -24,32 +24,48 @@ export class KycRepository extends IKycRepository {
...
+    return {
+      address: kycDbModel.address,
+      country: kycDbModel.country,
+      curp: kycDbModel.curp,
+      documentNumber: kycDbModel.documentNumber,
+      id: EntityId.fromString(kycDbModel.id),
+      subtype: kycDbModel.subtype,
+      userId: EntityId.fromString(kycDbModel.uid),
+    };
+  }
```
xeno097, line 37: this should be a function instead of a private method in the class

src/modules/user/infra/repositories/kyc.repository.ts
```diff
@@ -85,6 +101,10 @@ export class KycRepository extends IKycRepository {
 
   public async update(entity: KycEntity): Promise<KycEntity> {
     try {
+      if (!entity.id || !entity) {
+        throw new Error('Missing entity id or missing entity');
+      }
```
xeno097, line 106: this can be omitted if something like this can happen we need to make the current typescript rules more strict and also add input validations when receiving the data from metamap

src/utils/constants/index.ts
```diff
@@ -14,6 +14,11 @@ export class Constants {
...
+    REJECTED: 'FAILED',
+  };
```
xeno097, line 20: This should probably be an enum

src/modules/user/core/errors/Metamap/resource-not-found.error.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseError } from 'src/core/errors/base.error';
+
+export class KycNotFoundError extends BaseError {
+  constructor() {
+    super({
+      code: 'KYC_NOT_FOUND',
+      message: 'KYC data not found.',
+    });
+  }
+}
```
xeno097, line 10: You can reuse the `NotFoundError` instead of creating a new error only for the kyc
image

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,183 @@
...
+    private kycRequestRepository: IKycRequestRepository,
+    private constants: Constants
+  ) {}
+
+  private async getUserById(uid: string) {
+    const userIdEntity = EntityId.fromString(uid);
+    const user = await this.userRepository.getById(userIdEntity);
+    return user;
+  }
+
```
xeno097, line 30: remember to remove this in this case is not needed a function to wrap this logic you can directly use the `getById` method from the user repository where needed

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,183 @@
...
+  async createKycData(uid: string) {
+    const kycData = await this.metamapProvider.retrieveKycData(uid);
+    if (!kycData || !kycData.documents || kycData.documents.length === 0) {
+      throw new MissingKycDataError();
+    }
```
xeno097, line 35: In which scenario is this possible? Can we avoid having these fields set to undefined?

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,183 @@
...
+      const { resource, eventName } = body;
+
+      if (!resource || !eventName) {
+        console.log('Invalid request body');
+      }
+      const resourceData = await this.metamapProvider.getResouceData(resource);
+
+      let response: any;
+      if (eventName === 'verification_started') {
+        await this.updateUserVerified(resourceData.metadata.ui, 'PENDING');
```
xeno097, line 167: Following along with other conditions, shouldn't this be `resourceData.metadata.uid`? That is why correctly typing variables is important

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,183 @@
...
+    await this.updateUserVerified(uid, status);
+    await this.saveKycData(uid);
+
+    return { status: 200, message: exisitingResource ? 'RESOURCE_UPDATED' : 'RESOURCE_SAVED' };
+  }
+
+  async handleMetamapVerification(request: Request) {
+    try {
+      const body = await request.json();
+      const { resource, eventName } = body;
```
xeno097, line 158: This should be moved to the webhook class because you are handling HTTP request data directly when it should be done in the API layer.

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,183 @@
...
+      await this.kycRequestRepository.create(newResource);
+    } else {
+      exisitingResource.resourceId = resourceId;
+      await this.kycRequestRepository.update(exisitingResource);
+    }
+
+    await this.updateUserVerified(uid, status);
+    await this.saveKycData(uid);
+
+    return { status: 200, message: exisitingResource ? 'RESOURCE_UPDATED' : 'RESOURCE_SAVED' };
```
xeno097, line 152: This should be handled by the api layer as you are directly working with HTTP related data

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const url = `${this.metamap.apiUrl}/v2/verifications/${kycRequest.resourceId}`;
+    const token = this.getJWT();
+    const headers = {
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${token}`,
+      },
+    };
```
xeno097, line 66: You could probably refactor this into a method that takes a URL endpoint and the params/request body so you can avoid repeating this logic in every method with the only changing bit being the params and the url

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
```diff
@@ -0,0 +1,84 @@
...
+import { EntityId } from 'src/core/domain/entity/entity-id';
+import { KycNotFoundError } from 'src/modules/user/core/errors/Metamap/resource-not-found.error';
+import { IKycRequestRepository } from '../../repositories/kyc-request.repository';
+import { IKycRepository } from '../../repositories/kyc.repository';
+
+@Injectable()
+export class MetamapProvider {
+  private metamap: MetamapConfig;
+  private kycRepository: IKycRepository;
+  private kycRequestRepository: IKycRequestRepository;
```
xeno097, line 17: you are not initializing the repositories which will cause the app to crash when calling a method on any of those 2

xeno097 (changes requested):

src/modules/user/core/services/kyc.service.new.ts
DHurtado714-itesm: Are yo sure? because the file `kyc.repository.ts` works on the db communication layer. This file works con the kyc main logic, which combines repository and provider layer.

pellerano: Maybe not the entire file, but the class methods which only calls the repository feel redundant.

src/modules/user/enums/index.ts
```diff
@@ -0,0 +1,4 @@
+export enum MetamapStatus {
+  REVIEW_NEEDED = 'PENDING',
+  REJECTED = 'FAILED',
+}
```
xeno097, line 4: I think a `COMPLETED` status is missing here

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
```diff
@@ -0,0 +1,83 @@
...
+import { lastValueFrom, map } from 'rxjs';
+import { MetamapConfig } from 'src/core/config/app-config.interface';
+import { EnvConfigKey } from 'src/core/config/constants';
+import { EntityId } from 'src/core/domain/entity/entity-id';
+import { NotFoundError } from 'src/core/errors/not-found.error';
+import { IKycRequestRepository } from 'src/modules/user/core/infra/kyc-request.repository';
+import { IKycRepository } from 'src/modules/user/core/infra/kyc.repository';
+
+@Injectable()
+export class MetamapProvider {
```
xeno097, line 15: Remember to create an abstract class to hide the implementation of the concrete provider itself like I did for the CognitoProvider which is the implementation for the AuthProvider class

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
```diff
@@ -0,0 +1,83 @@
...
+  private metamap: MetamapConfig;
+  constructor(
+    private httpService: HttpService,
+    private configService: ConfigService,
+    private kycRepository: IKycRepository,
+    private kycRequestRepository: IKycRequestRepository
```
xeno097, line 21: Please add the `readonly` modifier to each one of the injected classes. `private readonly ...`

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
```diff
@@ -0,0 +1,83 @@
...
+      },
+    };
+
+    const request = this.httpService[method]<T>(url, body, headers);
+    return lastValueFrom(request.pipe(map((response: AxiosResponse<T>) => response.data)));
+  }
+
+  async completeVerification(request: Request): Promise<any> {
+    return request;
+  }
```
xeno097, line 42: Is this supposed to stay like this?

src/modules/user/api/webhook/metamap.webhook.ts
```diff
@@ -1,13 +1,13 @@
...
-import { Request } from 'express';
-import { MetamapService } from '../../infra/libs/metamap/metamap.service';
+import { KycService } from '../../core/services/kyc.service.new';
+import { MetamapBasePayload } from '../../core/services/types';
 
 @Controller('metamap')
-export class MetamapController {
-  constructor(private metamapService: MetamapService) {}
+export class MetamapWebhook {
+  constructor(private kycService: KycService) {}
```
xeno097, line 7: Please add the `readonly` modifier

src/modules/user/core/services/types.ts
```diff
@@ -8,3 +8,15 @@ export interface CompletedSignInResponseDto extends Omit<CompleteSignInResponseD
...
+  resource: string;
+  eventName: string;
+  flowId: string;
+  timestamp: string;
+  metadata: {
+    customTag?: string;
+    userId: number;
+    campaing: string;
+  };
+}
```
xeno097, line 22: Could you add documentation to the fields?

src/modules/user/core/services/types.ts
```diff
@@ -8,3 +8,15 @@ export interface CompletedSignInResponseDto extends Omit<CompleteSignInResponseD
 export interface DeleteUserAccountDto extends DeleteUserDto {
   id: string;
 }
+
+export interface MetamapBasePayload {
+  resource: string;
+  eventName: string;
```
xeno097, line 14: Considering that the documentation specifies exactly which events type could be sent by the API can you create an enum that describes those types? Also, I see that this is used in the controller so it shouldn't be placed here or you should create a new type that is valid only in the core layer

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
```diff
@@ -0,0 +1,83 @@
...
+  // async retreiveResourceDataWithUid(uid: string): Promise<any> {
+  //   const kycRequest = await this.kycRequestRepository.getByUserId(EntityId.fromString(uid));
+  //   if (!kycRequest) {
+  //     throw new NotFoundError();
+  //   }
+
+  //   const url = `${this.metamap.apiUrl}/v2/verifications/${kycRequest.resourceId}`;
+  //   return this.getResouceData('GET', url);
+
+  async getResouceData(resourceUrl: string): Promise<any> {
```
xeno097, line 79: Please remember to add a return type that is not any

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
```diff
@@ -0,0 +1,83 @@
...
+    return response.access_token;
+  }
+
+  async retrieveKycData(uid: string): Promise<any> {
+    const userIdEntity = EntityId.fromString(uid);
+    const kyc = await this.kycRepository.getByUserId(userIdEntity);
+    if (!kyc) throw new NotFoundError();
+    return kyc;
+  }
+
```
xeno097, line 69: This shouldn't be here as it uses the user repository

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,190 @@
...
+import { KycEntity } from '../domain/entities/kyc.entity';
+import { MissingKycDataError } from '../errors/Metamap/missing-kyc-data.error';
+import { IKycRequestRepository } from '../infra/kyc-request.repository';
+import { IKycRepository } from '../infra/kyc.repository';
+import { IUserRepository } from '../infra/user.repository';
+import { MetamapBasePayload } from './types';
+
+registerEnumType(MetamapStatus, {
+  name: 'MetamapStatus',
+});
```
xeno097, line 20: This should be placed in the API layer as it is something graphql API specific

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,190 @@
...
+      case MetamapStatus.REVIEW_NEEDED:
+        return MetamapStatus.REVIEW_NEEDED;
+      case MetamapStatus.REJECTED:
+        return MetamapStatus.REJECTED;
+      default:
+        return status.toUpperCase();
+    }
+  }
+
+  async updateUserVerified(uid: string, status: string) {
```
xeno097, line 108: Remember to add return types

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,190 @@
...
+});
+
+@Injectable()
+export class KycService {
+  constructor(
+    private metamapProvider: MetamapProvider,
+    private kycRepository: IKycRepository,
+    private userRepository: IUserRepository,
+    private kycRequestRepository: IKycRequestRepository,
+    private constants: Constants
```
xeno097, line 29: Please add the `readonly` modifier to these classes.

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,190 @@
...
+    await this.saveKycData(uid);
+    // TODO: notification service
+  }
+
+  async cancelKycStatus(uid: string): Promise<boolean> {
+    const userIdEntity = EntityId.fromString(uid);
+    const user = await this.userRepository.getById(userIdEntity);
+
+    if (user.kycStatus !== 'VERIFIED') {
+      user.kycStatus = null;
```
xeno097, line 126: Wouldn't it be better to have an unverified status for the user instead of setting it to null? Also I think we should throw an error if this method is called and the user is already verified

src/modules/user/api/webhook/metamap.webhook.ts
```diff
@@ -1,13 +1,13 @@
...
 
   @Post('/verification/complete')
-  completeVerification(@Req() request: Request) {
-    return this.metamapService.retrieveResourceData(request);
+  async completeVerification(@Req() request: MetamapBasePayload) {
```
xeno097, line 10: I don't think that this will work because that payload is in the  `body` property of the request type. You should use the `Body` decorator to extract the payload from the request

xeno097 (changes requested):

src/modules/user/enums/index.ts
DHurtado714-itesm: As it was on the previous implementation this was missing

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
DHurtado714-itesm: I really don't know, it was like this. I can check if it's necessary

src/modules/user/api/webhook/types.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @property {string} resource - The URL to the full resource information of the user's verification.
+ * @property {string} eventName - The name of the event in the verification life cycle.
+ * @property {string} flowId - The ID of the flow that the verification belongs to.
+ * @property {string} timestamp - The timestamp of when the webhook was sent.
+ * @property {Object} metadata - Custom values sent when initiating the verification process.
+ * @property {string} [metadata.customTag] - An optional custom tag for additional data.
+ * @property {number} metadata.userId - The user ID associated with the verification.
+ * @property {string} metadata.campaing - The campaign associated with the verification.
+ */
```
xeno097, line 15: I think that it would be better to move the description of each field above the corresponding field on the `MetamapBasePayload` interface so that when you hover over a field name the ide displays the field description and that way you don't need to repeat the name of the field because it is clear by just looking at the type definition

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,206 @@
...
+        await this.updateUserVerified(resourceData.metadata.uid, mapedStauts);
+      } else if (eventName === 'verification_expired') {
+        await this.cancelKycStatus(resourceData.metadata.uid);
+      } else if (eventName === 'verification_completed') {
+        const { status } = resourceData.identity;
+        if (!status) throw new NotFoundError();
+        const mapedStauts = mapToKycStatus(status);
+        await this.updateUserVerified(resourceData.metadata.uid, mapedStauts);
+        response = await this.saveResourceData(resourceData);
+      }
```
xeno097, line 199: Remember to replace hard coded event names and statuses with the enums that you created

xeno097 (changes requested):
I think that we are basically done with this only minor things remain and we can merge this

xeno097 (approved):

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,175 @@
...
+    currentKyc.country = country;
+    currentKyc.address = address;
+    currentKyc.curp = curp;
+    currentKyc.subtype = subtype ?? type;
+    currentKyc.documentNumber = documentNumber;
+    if (await this.kycRepository.update(currentKyc)) {
+      return { success: true, message: 'KYC data updated successfully' };
+    } else {
+      return { success: false, message: 'KYC data update failed' };
+    }
```
xeno097, line 82: This if is not needed because if the update fails an error will be thrown anyways the update method does not return a boolean, it returns the updated entity

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  async cancelKyc(uid: string): Promise<{ success: boolean; message: string }> {
+    const userIdEntity = EntityId.fromString(uid);
+    const user = await this.userRepository.getById(userIdEntity);
+
+    if (user.kycStatus === MetamapStatus.VERIFIED) {
+      throw new Error('Cannot cancel KYC status for an already verified user.');
+    }
+
+    if (user.kycStatus !== 'VERIFIED') {
```
xeno097, line 103: I see hardcoded strings here, also, if the user is already verified you should throw an error instead

src/modules/user/core/services/kyc.service.new.ts
```diff
@@ -0,0 +1,175 @@
...
+      user.kycStatus = MetamapStatus.UNVERIFIED;
+      await this.userRepository.update(user);
+      // TODO: delete kyc data
+      // await this.kycRepository.deleteByUserId(userIdEntity);
```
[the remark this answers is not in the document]
DHurtado714-itesm: I was debating if we should implements, delete kyc Data when the operation is expired. What do you think?

src/modules/user/enums/index.ts
```diff
@@ -0,0 +1,8 @@
+export enum MetamapStatus {
+  VERIFIED = 'VERIFIED',
+  PENDING = 'PENDING',
+  REVIEW_NEEDED = 'PENDING',
+  REJECTED = 'FAILED',
```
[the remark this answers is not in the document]
DHurtado714-itesm: Metamap has the status has `REJECTED`, but the DB has it as `FAILED`

src/modules/user/core/domain/value-objects/metamap.value-object.ts
```diff
@@ -0,0 +1,42 @@
...
+    super({
+      message: `Provided resourceId ${resourceId} is not valid`,
+    });
+  }
+}
+
+/**
+ * Wrapper type for a string that is a valid resourceId.
+ */
+export class MetamapResourceId extends BaseValueObject<string> {
```
xeno097, line 18: Could you please rename this `KycExternalId`? This also means that you should update the error name

src/modules/user/infra/repositories/kyc.repository.ts
```diff
@@ -17,44 +18,77 @@ export class KycRepository extends IKycRepository {
...
+        throw new NotFoundError(RESOURCE_NAME);
+      }
+
+      return kycEntityFromDbModel(kyc);
+    } catch (error) {
+      throw error;
+    }
+  }
+
+  public async getByResourceId(id: string): Promise<KycEntity> {
```
xeno097, line 53: Considering that we now have a dedicated type for the resource id could use it here?

src/modules/user/core/errors/metamap.error.ts
```diff
@@ -0,0 +1,19 @@
...
+  }
+}
+
+export class UserIsVerifiedError extends BaseError {
+  constructor(uid: string) {
+    super({
+      code: `USER${uid}_IS_VERIFIED`,
+      message: 'The user is already verified.',
+    });
+  }
```
xeno097, line 18: It would be better to leave the code as `VERIFIED_USER` or `USER_IS_VERIFIED` and update the error message to include the user's id. This way the front-end guys can handle the error without problems because the error code is predictable

src/modules/user/infra/libs/metamap/metamap.kyc.provider.ts
```diff
@@ -0,0 +1,73 @@
...
+import { MetamapApiResponse } from './types';
+
+@Injectable()
+export class MetamapProvider {
+  private metamap: MetamapConfig;
+  constructor(private readonly httpService: HttpService, private readonly configService: ConfigService) {
+    const metamapConfig = this.configService.getOrThrow<MetamapConfig>(EnvConfigKey.METAMAP);
+    if (!metamapConfig) {
+      throw new NotFoundError('METAMAP_CONFIG');
+    }
```
xeno097, line 18: you don't need to perform this check anymore here as the config service will throw an error for you if this is not set

src/modules/user/api/graphql/metamap.resolver.ts
```diff
@@ -2,15 +2,15 @@ import { UseGuards } from '@nestjs/common';
...
 import ContextType from 'src/infra/api/graphql/context-type';
 import { JwtAuthGuard } from 'src/infra/api/guards/jwt-auth.guard';
-import { MetamapService } from '../../infra/libs/metamap/metamap.service';
+import { KycService } from '../../core/services/kyc.service.new';
 
 @Resolver(() => String)
-export class MetamapResolver {
-  constructor(private metamapService: MetamapService) {}
+export class KycResolver {
+  constructor(private readonly metamapService: KycService) {}
```
xeno097, line 9: please rename the variable to `kycService`

src/modules/user/core/services/kyc.service.new.ts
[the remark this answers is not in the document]
xeno097: if a KYC is expired a new one is created or the same can be restarted?

xeno097 (changes requested):

xeno097 (approved):

xeno097 (approved):

PR #136, opened 2023-12-19
fix(cognito): is user auth in cognito flow

DHurtado714-itesm:
Query in graphql for IsUserAuthenticated fucntion but in cognito flow.


pellerano (approved):

src/modules/user/infra/libs/cognito/cognito.auth.provider.ts
```diff
@@ -430,4 +431,22 @@ export class AwsCognitoAuthProvider extends IAuthProvider {
...
+        Username: dto.accessToken,
+      });
+
+      const response = await this.cognitoProvider.send(command);
+      if (response.$metadata.httpStatusCode === 200) {
+        return true;
+      }
+      return false;
+    } catch (error) {
+      console.error(error);
```
xeno097, line 448: you can avoid logging the error here because the global catcher will take care of that

src/modules/user/core/infra/auth-provider.ts
```diff
@@ -140,6 +140,10 @@ export interface DeleteUserDto {
   accessToken: string;
 }
 
+export interface isUserAuthenticatedDto {
```
xeno097, line 143: The name should be `IsUserAuthenticated` ?

src/modules/user/api/graphql/auth.resolver.new.ts
```diff
@@ -181,4 +181,10 @@ export class NewAuthResolver {
       USDC: 0,
     };
   }
+
+  @Query(() => Boolean, { description: 'Returns currently authenticated user session info (cognito).' })
+  @UseGuards(JwtAuthGuard)
+  async isUserAuthenticated(@JwtPayload(true) jwtPayload: JwtTokenPayloadV1): Promise<boolean> {
```
xeno097, line 187: This should be renamed because it shadows the currently existing `isUserAuthenticated` query that supports the amplify auth flow

xeno097 (changes requested):
I have question. Do we really need a `isUserAuthenticated` with the new flow considering that if the user is not, the backend will either inform the app that the token has expired or it is invalid if none is provided with a request?

AlbertoMorillo (approved):

xeno097:
I thnk that we all agree that this can be closed now

PR #153, opened 2024-01-18
Notification module refactor part 2

DHurtado714-itesm:
Implemented the base use-cases for notifications module


src/modules/notifications/core/use-cases/deposit-notfication.use-case.ts
```diff
@@ -0,0 +1,31 @@
+import { Injectable } from '@nestjs/common';
+import { IUseCase } from 'src/core/domain/use-cases/use-case.interface';
+import { FirebaseNotificationsProvider } from '../../infra/lib/firebase/firebase.notifications.provider';
+import { NotificationMessage, orderNotificationMessages } from '../../notification-message.interface';
+import { DepositNotificationUseCaseDTO } from '../dto/notification-use-case.dto';
+import { NotificationTransactionStatus } from '../enums/notification.enums';
+
+@Injectable()
+export class DepositUseCase implements IUseCase<DepositNotificationUseCaseDTO, Promise<void>> {
+  constructor(private firebaseNotificationProvider: FirebaseNotificationsProvider) {}
```
xeno097, line 10: please add the `readonly` modifier after `private` and use the interface instead of the concrete implementation `FirebaseNotificationsProvider` if don't have one create a `INotificationProvider` interface as we did for the repositories.

src/modules/notifications/core/use-cases/deposit-notfication.use-case.ts
```diff
@@ -0,0 +1,31 @@
...
+        notificationMessage = orderNotificationMessages.depositConfirmedMessage;
+        break;
+      case NotificationTransactionStatus.PENDING:
+        notificationMessage = orderNotificationMessages.depositPendingMessage;
+        break;
+      case NotificationTransactionStatus.FAILED:
+        notificationMessage = orderNotificationMessages.depositFailedMessage;
+        break;
+      default:
+        throw new Error('Invalid deposit status');
```
xeno097, line 26: Please create a custom error using the provided classes

xeno097 (changes requested):
Please add unit tests to the use case

src/modules/notifications/infra/lib/firebase/firebase.notifications.provider.ts
```diff
@@ -4,10 +4,11 @@ import { ConfigService } from '@nestjs/config';
 import { firstValueFrom } from 'rxjs';
 import { FirebaseConfig } from 'src/core/config/app-config.interface';
 import { EnvConfigKey } from 'src/core/config/constants';
+import { INotificationsProvider } from 'src/modules/notifications/core/infra/push-notifications.provider';
 import { NotificationMessage } from 'src/modules/notifications/notification-message.interface';
 
 @Injectable()
-export class FirebaseNotificationsProvider {
+export class FirebaseNotificationsProvider implements INotificationsProvider {
   private firebaseConfig: FirebaseConfig;
```
xeno097, line 12: It should be `extends` instead of implements as it is inheriting from an abstract class

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  it('should throw InvalidOrderStatusError for an invalid status', async () => {
+    const dto: DepositNotificationUseCaseDTO = {
+      txnType: IOrdeType.DEPOSIT,
+      status: 'INVALID_STATUS' as NotificationTransactionStatus,
+      deviceToken: 'device_token_example', // TODO: change this
+    };
+
+    await expect(depositUseCase.execute(dto)).rejects.toThrow(InvalidOrderStatusError);
+  });
```
xeno097, line 52: Please use the same test structure used in the other unit tests . See [this](https://dev.to/coderjay06/the-three-a-s-of-unit-testing-b22) article for what I'm talking about

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    await depositUseCase.execute(dto);
+
+    expect(notificationsProvider.sendPushNotification).toHaveBeenCalledWith(
+      orderNotificationMessages.depositConfirmedMessage,
+      dto.deviceToken
+    );
+  });
+
+  it('should throw InvalidOrderStatusError for an invalid status', async () => {
```
xeno097, line 44: you can use the class to get the name instead of manually typing it so that if for whatever reason we change the name of the error it is automatically changed here. Example:
image

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { NotificationTransactionStatus } from '../../enums/notification.enums';
+import { InvalidOrderStatusError } from '../../errors/invalid-status.error';
+import { INotificationsProvider } from '../../infra/push-notifications.provider';
+import { DepositUseCase } from '../deposit-notfication.use-case';
+
+enum IOrdeType {
+  DEPOSIT = 'DEPOSIT',
+}
+
+describe('DepositUseCase', () => {
```
xeno097, line 13: Same comment as below you can directly use the class and take its name

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  let depositUseCase: DepositUseCase;
+  let notificationsProvider: INotificationsProvider;
+
+  beforeEach(async () => {
+    notificationsProvider = {
+      sendPushNotification: jest.fn(),
+    };
```
xeno097, line 20: Please create a mock function in a separate file as done in the other tests. Example:
image

src/modules/notifications/core/errors/invalid-status.error.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseError } from 'src/core/errors/base.error';
+
+export class InvalidOrderStatusError extends BaseError {
+  constructor(type = 'ORDER') {
+    super({
+      code: `INVALID_${type.toUpperCase()}_STATUS_ERROR`,
+      message: `Invalid ${type.toLowerCase()} status, please provide a valid one.`,
+    });
+  }
+}
```
xeno097, line 10: I think that this should `extend` from `BaseInternalError` because is not an error caused by a user input but instead it is caused by a programming bug. It would also be a good idea to add which ones are the valid values from the enum.

src/modules/notifications/core/use-cases/deposit-notfication.use-case.ts
```diff
@@ -0,0 +1,32 @@
+import { Injectable } from '@nestjs/common';
+import { IUseCase } from 'src/core/domain/use-cases/use-case.interface';
+import { FirebaseNotificationsProvider } from '../../infra/lib/firebase/firebase.notifications.provider';
+import { NotificationMessage, orderNotificationMessages } from '../../notification-message.interface';
+import { DepositNotificationUseCaseDTO } from '../dto/notification-use-case.dto';
+import { NotificationTransactionStatus } from '../enums/notification.enums';
+import { InvalidOrderStatusError } from '../errors/invalid-status.error';
+
+@Injectable()
+export class DepositUseCase implements IUseCase<DepositNotificationUseCaseDTO, Promise<void>> {
```
xeno097, line 10: you can simply put `void` as the return type without the `Promise`

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { orderNotificationMessages } from 'src/modules/notifications/notification-message.interface';
+import { DepositNotificationUseCaseDTO } from '../../dto/notification-use-case.dto';
+import { NotificationTransactionStatus } from '../../enums/notification.enums';
+import { InvalidOrderStatusError } from '../../errors/invalid-status.error';
+import { INotificationsProvider } from '../../infra/push-notifications.provider';
+import { DepositUseCase } from '../deposit-notfication.use-case';
+
+enum IOrdeType {
+  DEPOSIT = 'DEPOSIT',
+}
```
xeno097, line 11: Do we really need an enum for this?

xeno097 (changes requested):

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
DHurtado714-itesm: I need to add more Order types, but yes, for not using any hardcoded strings

pellerano (dismissed):

src/modules/notifications/infra/lib/firebase/firebase.notifications.provider.ts
```diff
@@ -4,30 +4,33 @@ import { ConfigService } from '@nestjs/config';
...
-        Authorization: `key=${firebaseKey}`,
+        Authorization: `key=${this.firebaseConfig.cloudMessagingKey}`,
         'Content-Type': 'application/json',
       },
     };
   }
 
-  // TODO: Check Promise, maybe use a interface for deviceToken, custom errors
+  // TODO: custom errors
   async sendPushNotification(notification: NotificationMessage, deviceToken: string): Promise<void> {
```
xeno097, line 28: Considering that you created a `DeviceToken` type, use that instead of a plain string so you don't need to verify that the string is defined.

src/modules/notifications/infra/lib/firebase/firebase.notifications.provider.ts
```diff
@@ -4,30 +4,33 @@ import { ConfigService } from '@nestjs/config';
...
-    if (!notification.title || !notification.message || !deviceToken) {
-      throw new Error('Notification must have a title and a message');
+    // TODO: Implement device token validator
+    if (!notification || !deviceToken) {
+      throw new Error('Notification or deviceToken not provided');
     }
```
xeno097, line 32: You can remove this error checking as neither of the function parameters can be undefined

src/modules/notifications/core/dto/validator/device-token.validator.ts
```diff
@@ -0,0 +1,43 @@
...
+   * Regex and text length is implemented as a validator
+   */
+  private static isValidDeviceToken(deviceToken: string): boolean {
+    const maxLength = 256;
+    const minLength = 64;
+    const regex = /^[A-Za-z09-_:]+$/;
+
+    return deviceToken.length >= minLength && deviceToken.length <= maxLength && regex.test(deviceToken);
+  }
+}
```
xeno097, line 43: This file should be placed under the folder `src/modules/notifications/core/domain/value-objects` as you can see from the structure of the `user` module

src/modules/notifications/core/services/notification.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  public async sendWithdrawNotification(dto: WithdrawNotificationUseCaseDTO): Promise<void> {
+    await this.withdrawUseCase.execute(dto);
+  }
+
+  public async sendMetamapNotification(dto: MetamapNotificationUseCaseDTO): Promise<void> {
+    await this.metamapUseCase.execute(dto);
+  }
+}
```
xeno097, line 30: This file can be removed entirely as use cases should not be used in service classes but instead should be invoked directly by their callers being it a controller or any other public API.

src/modules/notifications/core/enums/notification.enums.ts
```diff
@@ -0,0 +1,24 @@
...
+  PENDING = 2,
+  /**
+   * We dont have a notification for this status
+   */
+  EXPIRED = 3,
+  /**
+   * We dont have a notification for this status
+   */
+  CANCELLED = 4,
+}
```
xeno097, line 13: considering that this enum is new I'd prefer to use a string instead of a number for the internal value as you did it for the `OrderTypeEnum`

src/modules/notifications/core/enums/notification.enums.ts
```diff
@@ -0,0 +1,24 @@
...
+export enum OrderTypeEnum {
+  DEPOSIT = 'DEPOSIT',
+  WITHDRAW = 'WITHDRAW',
+}
+
+export enum MetamapNotificationStatus {
+  VERIFIED = 'VERIFIED',
+  PENDING = 'PENDING',
+  FAILED = 'FAILED',
+}
```
xeno097, line 24: We should reach an agreement on enum naming because sometimes you add the `Enum` suffix and sometimes you don't which makes code inconsistent. Considering that the majority of the code does not use the `Enum` suffix, I'd remove it here.

src/modules/notifications/core/infra/__test__/mocks/notifications.provider.mock.ts
```diff
@@ -0,0 +1,7 @@
+import { INotificationsProvider } from '../../push-notifications.provider';
+
+export const sendPushNotificationMock = (): jest.Mocked<INotificationsProvider> => {
```
xeno097, line 3: Please rename the function to `getNotificationProviderMock` to follow the convention for naming functions that create mocked classes.

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -0,0 +1,120 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { FirebaseNotificationsProvider } from 'src/modules/notifications/infra/lib/firebase/firebase.notifications.provider';
+import { orderNotificationMessages } from 'src/modules/notifications/notification-message.interface';
+import { DepositNotificationUseCaseDTO } from '../../dto/notification-use-case.dto';
+import { InvalidDeviceTokenError } from '../../dto/validator/device-token.validator';
+import { NotificationTransactionStatus, OrderTypeEnum } from '../../enums/notification.enums';
+import { sendPushNotificationMock } from '../../infra/__test__/mocks/notifications.provider.mock';
+import { DepositNotificationUseCase } from '../deposit-notfication.use-case';
+
+const DEVICE_TOKEN = '';
```
xeno097, line 10: Should this be empty?

src/modules/notifications/core/use-cases/test/metamap-use-case.test.ts
```diff
@@ -0,0 +1,108 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { FirebaseNotificationsProvider } from 'src/modules/notifications/infra/lib/firebase/firebase.notifications.provider';
+import { orderNotificationMessages } from 'src/modules/notifications/notification-message.interface';
+import { MetamapNotificationUseCaseDTO } from '../../dto/notification-use-case.dto';
+import { InvalidDeviceTokenError } from '../../dto/validator/device-token.validator';
+import { MetamapNotificationStatus } from '../../enums/notification.enums';
+import { sendPushNotificationMock } from '../../infra/__test__/mocks/notifications.provider.mock';
+import { MetamapNotificationUseCase } from '../metamap-notification.use-case';
+
+const DEVICE_TOKEN = '';
```
xeno097, line 10: x2

src/modules/notifications/core/use-cases/test/withdraw-use-case.test.ts
```diff
@@ -0,0 +1,117 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { FirebaseNotificationsProvider } from 'src/modules/notifications/infra/lib/firebase/firebase.notifications.provider';
+import { orderNotificationMessages } from 'src/modules/notifications/notification-message.interface';
+import { WithdrawNotificationUseCaseDTO } from '../../dto/notification-use-case.dto';
+import { InvalidDeviceTokenError } from '../../dto/validator/device-token.validator';
+import { NotificationTransactionStatus, OrderTypeEnum } from '../../enums/notification.enums';
+import { sendPushNotificationMock } from '../../infra/__test__/mocks/notifications.provider.mock';
+import { WithdrawNotificationUseCase } from '../withdraw-notification.use-case';
+
+const DEVICE_TOKEN = '';
```
xeno097, line 10: x3

src/modules/notifications/core/dto/notification-use-case.dto.ts
```diff
@@ -0,0 +1,62 @@
+import { SupportedCryptoCurrency } from 'src/utils/enums';
+import { MetamapNotificationStatus, NotificationTransactionStatus } from '../enums/notification.enums';
+
+export interface DepositNotificationUseCaseDTO {
```
xeno097, line 4: Please rename this to `DepositNotificationUseCaseDto` the same applies to all the other types that have the `DTO` suffix.

xeno097 (changes requested):
One final note is that my preference would be for use cases Dtos to be placed in the same file as their use case. Example:

image

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
DHurtado714-itesm: It depends on the dev who is testing, each one should use their own DEVICE_TOKEN

src/modules/notifications/core/enums/notification.enums.ts
DHurtado714-itesm: Here `src/utils/enums/index.ts` we have a enum which specifies the Transaction type for the database, I am using the same logic

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -0,0 +1,103 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { FirebaseNotificationProvider } from 'src/modules/notifications/infra/lib/firebase/firebase.notifications.provider';
+import { orderNotificationMessages } from 'src/modules/notifications/notification-message.interface';
+import { DeviceToken, InvalidDeviceTokenError } from '../../domain/value-objects/device-token.validator';
+import { NotificationTransactionStatus, OrderType } from '../../enums/notification.enums';
+import { getNotificationProviderMock } from '../../infra/__test__/mocks/notifications.provider.mock';
+import { DepositNotificationUseCase, DepositNotificationUseCaseDto } from '../deposit-notfication.use-case';
+
+// TODD: check this with @Sebastiano
```
xeno097, line 9: I think that you can remove these comments now

src/modules/notifications/core/enums/notification.enums.ts
xeno097, line 13: I think you forgot to update this here

xeno097 (changes requested):

xeno097 (dismissed):

PR #160, opened 2024-01-24
Add obfuscated destination to forgot password.

pellerano:
### Changelog

This PR implements the obfuscated telephone number or email on forgot password endpoint.

- Implements and tests the obfuscated telephone number or email on forgot password.

Examples return objects:
image
image


src/modules/user/api/graphql/object-types/auth.object-types.ts
```diff
@@ -64,7 +64,10 @@ export class BaseDeliveryMediumResponse {
 }
 
 @ObjectType()
-export class ForgotPasswordResponse extends BaseDeliveryMediumResponse {}
+export class ForgotPasswordResponse extends BaseDeliveryMediumResponse {
+  @Field(() => String)
+  destination?: string;
```
xeno097, line 69: This can't be nullable remove the `?` please

src/modules/user/core/infra/__tests__/mocks/auth-provider.mock.ts
```diff
@@ -8,7 +8,12 @@ export const getAuthProviderMock = (): jest.Mocked<IAuthProvider> => {
     completeSignIn: jest.fn(),
     confirmSignUp: jest.fn(),
     deleteUser: jest.fn(),
-    forgotPassword: jest.fn(),
+    forgotPassword: jest.fn(
+      (_: ForgotPasswordDto) =>
+        new Promise<ForgotPasswordResponseDto>(resolve => {
+          resolve({ deliveryMedium: CodeDeliveryMedium.SMS, destination: '******0000' });
+        })
+    ),
```
xeno097, line 16: Please use the `mockImplementation` method or `mockResolvedValueOnce` method in the test where you need this. There are examples in other unit tests of how to mock a return value. 
Let's try and avoid forcing a mock to have a predetermined return value or mocked implementation

src/modules/user/core/infra/auth-provider.ts
```diff
@@ -139,6 +139,7 @@ export interface ForgotPasswordDto {
 
 export interface ForgotPasswordResponseDto {
   deliveryMedium: CodeDeliveryMedium;
+  destination?: string;
```
xeno097, line 142: Please remove the `?`

src/modules/user/core/services/__tests__/auth.service.spec.ts
```diff
@@ -386,11 +386,13 @@ describe(NewAuthService.name, () => {
       };
 
       // Act
-      await authService.forgotPassword(dto);
+      const response = await authService.forgotPassword(dto);
+      console.log(response);
```
xeno097, line 390: Please remove this console log

src/modules/user/infra/libs/cognito/cognito.auth.provider.ts
```diff
@@ -737,6 +737,7 @@ export class AwsCognitoAuthProvider extends IAuthProvider {
 
       return {
         deliveryMedium: codeDeliveryMediumFromStr(CodeDeliveryDetails.DeliveryMedium),
+        destination: CodeDeliveryDetails.Destination,
```
xeno097, line 740: If the `Destination` field is possibly undefined from the cognito response then as a failsafe in the service method that calls this one, get the user from the database and assign the telephone number or the email based on the value of `deliveryMedium`. Alternatively, verify if the `Destination` field is set like it is done for the `DeliveryMedium` field as both should be defined if Cognito processed the request correctly. Example:
image

See the SDK docs for more details on this https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/cognito-identity-provider/command/ForgotPasswordCommand/

xeno097 (changes requested):

src/modules/user/core/infra/__tests__/mocks/auth-provider.mock.ts
```diff
@@ -8,7 +8,9 @@ export const getAuthProviderMock = (): jest.Mocked<IAuthProvider> => {
     completeSignIn: jest.fn(),
     confirmSignUp: jest.fn(),
     deleteUser: jest.fn(),
-    forgotPassword: jest.fn(),
+    forgotPassword: jest
+      .fn()
+      .mockResolvedValueOnce({ deliveryMedium: CodeDeliveryMedium.SMS, destination: '******0000' }),
```
xeno097, line 13: Please I already asked you to do this in the test case where you need it instead of forcing this on the mock itself when it is created

src/modules/user/core/services/__tests__/auth.service.spec.ts
```diff
@@ -386,11 +386,12 @@ describe(NewAuthService.name, () => {
       };
 
       // Act
-      await authService.forgotPassword(dto);
+      const response = await authService.forgotPassword(dto);
 
       // Assert
       expect(authProvider.forgotPassword).toHaveBeenCalledTimes(1);
       expect(authProvider.forgotPassword).toHaveBeenCalledWith(dto);
+      expect(response?.destination).toContain('*');
```
xeno097, line 394: The `?` is not needed anymore as this field can't be undefined now

xeno097 (changes requested):

xeno097 (approved):

PR #156, opened 2024-01-20
feat: implement the `CompleteBankDepositUseCase`

xeno097:
### Changelog

This PR implements the `CompleteBankDepositUseCase` which is going to deprecate the `releaseFunds` operation used by the backoffice.

- Implements and tests the `CompleteBankDepositUseCase`
- Adds the `masterWallet` method on the `IBlockchainProvider` (if you have a better name please suggest one)
- Defines the `BankDepositEntity` to describe transactions that are bank deposits
- Implements the `BankDepositRepository`
-  

Notes:
- This PR should be merged after #155 as it builds upon it

PR #162, opened 2024-01-29
Feat/OINK-1356: Update the CancelTransactionStatusUseCase

pellerano:
This PR updates the current implementation of the CancelTransactionStatusUseCase because it does not consider all the edge cases and potentially allows any user to delete a transaction.


DHurtado714-itesm (dismissed):

src/modules/transaction/core/use-cases/to-review/cancel-transaction.use-case.ts
```diff
@@ -1,16 +1,36 @@
...
+  }
+
   public async execute(input: CancelTransactionDto) {
     const transaction = await this.transactionRepository.getById(EntityId.fromString(input.transactionId));
+
+    // Check if transaction belongs to currently logged in user and throw error if it doesn't.
+    if (transaction.userId?.inner !== EntityId.fromString(input.userId).inner) throw new TransactionOwnershipError();
+
+    // Only change status if status is not COMPLETED/FAILED/CANCELLED/EXPIRED
+    if (this.isTransactionStatusChangeable(transaction.status)) return false;
```
xeno097, line 32: You should change this here as you are overcomplicating the logic. If a transaction can be canceled only when the status is `TransactionStatus.PENDING` then simply compare `transaction.status === TransactionStatus.PENDING`, also instead of simply returning false you should return an error as it is more meaningful than simply returning false which does not tell anything meaningful to the user about the result of the operation.

src/modules/transaction/core/use-cases/to-review/cancel-transaction.use-case.ts
```diff
@@ -1,16 +1,36 @@
...
+      TransactionStatus.EXPIRED,
+      TransactionStatus.IN_PROGRESS,
+    ].includes(status);
+  }
+
   public async execute(input: CancelTransactionDto) {
     const transaction = await this.transactionRepository.getById(EntityId.fromString(input.transactionId));
+
+    // Check if transaction belongs to currently logged in user and throw error if it doesn't.
+    if (transaction.userId?.inner !== EntityId.fromString(input.userId).inner) throw new TransactionOwnershipError();
```
xeno097, line 29: I would prefer that you return a NotFound error instead this way we don't inform the user that the transaction exists

src/modules/transaction/core/use-cases/__tests__/cancel-transaction.use-case.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  describe(CancelTransactionStatusUseCase.prototype.execute, () => {
+    it(`should throw an ${TransactionOwnershipError.name} if the user wanting to cancel a transaction is not the owner of said transaction`, async () => {
+      // Arrange
+      const { transactionRepository, cancelTransactionStatusUseCase } = await setup();
+      transactionRepository.getById.mockReturnValueOnce(transactionMock);
+
+      const dto: CancelTransactionDto = {
+        transactionId: EntityId.new().inner,
+        userId: `${EntityId.new().inner}`,
```
xeno097, line 45: you don't need to use a template string here, simply use `EntityId.new().inner`

src/modules/transaction/core/use-cases/__tests__/cancel-transaction.use-case.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      const { transactionRepository, cancelTransactionStatusUseCase } = await setup();
+
+      transactionRepository.getById.mockReturnValueOnce({
+        ...transactionMock,
+        status: TransactionStatus.CANCELLED,
+      });
+
+      const dto: CancelTransactionDto = {
+        transactionId: EntityId.new().inner,
+        userId: `${expectedUserId.inner}`,
```
xeno097, line 66: same here

src/modules/transaction/core/use-cases/__tests__/cancel-transaction.use-case.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    it(`should successfully process a transaction cancelation`, async () => {
+      // Arrange
+      const { transactionRepository, cancelTransactionStatusUseCase } = await setup();
+
+      transactionRepository.getById.mockReturnValueOnce(transactionMock);
+
+      const dto: CancelTransactionDto = {
+        transactionId: EntityId.new().inner,
+        userId: `${expectedUserId.inner}`,
```
xeno097, line 85: same here

src/modules/transaction/api/graphql/transactions.resolver.new.ts
```diff
@@ -108,10 +111,13 @@ export class TransactionResolver {
...
 
-  // @Mutation(() => Boolean)
-  // public async cancelTransaction(@Args('input') input: CancelTransactionInput) {
-  //   return this.cancelTransactionUseCase.execute(input);
-  // }
+  @Mutation(() => Boolean)
+  public async cancelTransaction(
+    @JwtPayload(true) jwtPayload: JwtTokenPayloadV1,
+    @Args('input') input: CancelTransactionInput
+  ) {
```
xeno097, line 118: It would help if you validated the user input. You can see examples above in the other operations. also, it would be best if you changed the name of this mutation to `cancelAuthenticatedUserTransaction` because this operation cancels a transaction that belongs to the currently authenticated user. Also please add documentation about this mutation and which operation from the old logic replaces.

src/modules/transaction/core/errors/transaction-ownership-error.error.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export class TransactionOwnershipError extends BaseError {
+  constructor() {
+    super({
+      code: 'TRANSACTION_OWNERSHIP_ERROR',
+      message:
+        'Unauthorized Transaction Access: User requesting transaction modification lacks permission to perform this action.',
+    });
+  }
+}
```
xeno097, line 11: You can  delete this after performing the changes

src/modules/transaction/core/use-cases/__tests__/cancel-transaction.use-case.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+import { CancelTransactionDto, CancelTransactionStatusUseCase } from '../to-review/cancel-transaction.use-case';
+
+describe(CancelTransactionStatusUseCase.name, () => {
+  const expectedUserId = EntityId.new();
+  const transactionMock: TransactionEntity = {
+    id: EntityId.new(),
+    createdAt: new Date(),
+    sourceAsset: SupportedFiatCurrency.DOP,
+    status: TransactionStatus.PENDING,
+    type: 'DEPOSIT',
```
xeno097, line 15: Please use the `TransactionType` enum instead of the hardcoded value

src/modules/transaction/core/use-cases/__tests__/cancel-transaction.use-case.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        transactionId: EntityId.new().inner,
+        userId: `${expectedUserId.inner}`,
+      };
+
+      // Act
+      const promise = await cancelTransactionStatusUseCase.execute(dto);
+
+      // Assert
+      expect(transactionRepository.getById).toHaveBeenCalledTimes(1);
+      expect(promise).toBe(true);
```
xeno097, line 93: you should assert that when the transaction was sent to the repository it did have the status set to `TransactionStatus.Cancelled`

xeno097 (changes requested):

src/modules/transaction/api/graphql/input-types/transaction.input-types.ts
```diff
@@ -17,3 +17,9 @@ export class P2pTransferGraphQLInput {
   @Field(() => ID)
   receiverId: string;
 }
+
+@InputType('CancelTransactionInput')
+export class CancelTransactionGraphQLInput {
+  @Field()
```
xeno097, line 23: Please add the `ID` type here

src/modules/transaction/core/use-cases/to-review/cancel-transaction.use-case.ts
```diff
@@ -1,18 +1,40 @@
...
+  implements IUseCase<ExecuteCancelTransactionUseCaseDto, CancelTransactionDto>
+{
   constructor(private readonly transactionRepository: ITransactionRepository) {}
 
-  public async execute(input: CancelTransactionDto) {
-    const transaction = await this.transactionRepository.getById(EntityId.fromString(input.transactionId));
+  public async execute(dto: ExecuteCancelTransactionUseCaseDto): Promise<CancelTransactionDto> {
+    const transaction = await this.transactionRepository.getById(EntityId.fromString(dto.transactionId));
+
+    if (transaction.userId?.inner !== EntityId.fromString(dto.userId).inner) throw new NotFoundError();
```
xeno097, line 32: you can simplify the statement to `if (transaction.userId !== EntityId.fromString(dto.userId))`

src/modules/transaction/core/use-cases/to-review/cancel-transaction.use-case.ts
```diff
@@ -1,18 +1,40 @@
...
+
```
xeno097, line 33: I think that we should also verify that the transaction type is `Deposit` because it does not make sense to cancel a blockchain transaction and it shouldn't even be allowed

src/modules/transaction/core/errors/transaction-not-cancelable.error.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseError } from 'src/core/errors/base.error';
+
+export class TransactionNotCancelableError extends BaseError {
+  constructor() {
+    super({
+      code: 'TRANSACTION_NOT_CANCELABLE_ERROR',
+      message: `The transaction cannot be cancelled due to its current status.`,
+    });
+  }
+}
```
xeno097, line 10: After #156 is merged please use this error instead:
image

xeno097 (changes requested):

xeno097 (dismissed):

xeno097:
Hey @pellerano now that #156 has been merged please replace the `TransactionNotCancelableError` with `InvalidTransactionStateTransitionError` and also solve the merge conflicts in the PR

src/modules/transaction/core/use-cases/to-review/cancel-transaction.use-case.ts
```diff
@@ -1,18 +1,44 @@
...
+    const transaction = await this.transactionRepository.getById(EntityId.fromString(dto.transactionId));
+
+    if (transaction.userId?.inner !== dto.userId) throw new NotFoundError();
+
+    if (transaction.status !== TransactionStatus.PENDING)
+      throw new InvalidTransactionStateTransitionError(transaction.status, TransactionStatus.CANCELLED);
+
+    if (transaction.type !== TransactionType.DEPOSIT)
+      throw new InvalidTransactionStateTransitionError(transaction.status, TransactionStatus.CANCELLED);
+
```
xeno097, line 39: I think that this should be a different error something that explains that you cannot cancel a transaction that is not a bank deposit

xeno097 (changes requested):

src/modules/transaction/core/use-cases/complete-user-bank-deposit.use-case.ts
```diff
@@ -21,6 +21,15 @@ export interface CompleteUserBankDepositResponseDto {
   token: SupportedCryptoCurrency;
 }
 
+export class InvalidTransactionCancellationError extends BaseError {
+  constructor(type: TransactionType) {
+    super({
+      code: 'INVALID_TRANSACTION_STATE_TRANSITION_ERROR',
```
xeno097, line 27: The error code is wrong here. Please update it to reflect the exact error that is happening.

xeno097 (changes requested):
Additionally just to finally close this PR can you mark as deprecated the `cancelTransaction` mutation in the old `TransactionResolver`?

src/modules/transaction/core/use-cases/complete-user-bank-deposit.use-case.ts
```diff
@@ -21,6 +21,15 @@ export interface CompleteUserBankDepositResponseDto {
   token: SupportedCryptoCurrency;
 }
 
+export class InvalidTransactionCancellationError extends BaseError {
+  constructor(type: TransactionType) {
+    super({
+      code: 'INVALID_TRANSACTION_CANCELLATION_ERROR',
+      message: `Invalid transaction state transition cannot update transaction status of transaction type: ${type}}`,
```
xeno097, line 28: Please update the error message to `Invalid transaction cancellation request. Cannot cancel transaction with status ${type}`

xeno097 (changes requested):

xeno097 (approved):

PR #163, opened 2024-02-01
feat: user wallet info db model

xeno097:
### Changelog

This PR implements the required changes to start the DFNS integration.

- Implements the `getAuthenticatedUserBalance` and the `getUserBalanceByWalletAddress`  that deprecate the `getWalletBalance` and `getBalance` queries respectively.
- Defines the `IBlockchainWalletPublicApi` and its concrete implementation to allow other modules to call its methods
- Implemented the `WalletId` value object to abstract a string that is a valid wallet id
- Adds the `create` and `getByBlockchainAddress` methods on the  `IUserBlockchainWalletInfoRepository`
- Adds `id` method on the `IBlockchainWallet`
- Implements a new `WalletService`
- Adds a new `UserBlockchainWalletInfoDBModel` (this will require a new table in the database)
- Updates the new `UserService` to use the `IBlockchainWalletPublicApi`


src/modules/transaction/api/graphql/wallet.resolver.new.ts
```diff
@@ -0,0 +1,45 @@
...
+    @JwtPayload(true) jwtPayload: JwtTokenPayloadV1
+  ): Promise<GetUserBlockchainWalletBalanceGraphQLResponse> {
+    const balance = await this.walletsService.getWalletBalanceByUserId(jwtPayload.userId);
+
+    return {
+      total: balance,
+    };
+  }
+
+  @AdminPanelOnly()
```
DHurtado714-itesm, line 31: Are we including the balance in the Panel?

xeno097: Honestly, I don't know but between the old wallet operations, there was a query that allowed anyone to query a wallet balance by providing a blockchain address. As it wasn't used in the app i assumed it was for the panel

src/modules/transaction/infra/database/postgres/models/user-blockchain-wallet-info.model.ts
DHurtado714-itesm, line 1: are we going to use snake_case or CamelCase

src/modules/transaction/api/graphql/wallet.resolver.new.ts
DHurtado714-itesm: I guess

DHurtado714-itesm (approved):

PR #168, opened 2024-02-08
Feat/notification refactor 3

DHurtado714-itesm:
## Describe
1. Implement a public api inside Notifications Module, so other modules can use the service.
2. Created a repository to fetch user `deviceToken` from the DB
3. Refactor some code to integrate the repository
4. Include in the BankDepositUseCase and MetamapService, the notification service
5. Re-do 1 test case, PLEASE check it before continuing with the next 2
6. Pending the withdraw notification, the use case is not implemented


src/modules/transaction/infra/mappers/notification-status.mapper.ts
```diff
@@ -0,0 +1,21 @@
...
+    case TransactionStatus.PENDING:
+      return NotificationTransactionStatus.PENDING;
+    case TransactionStatus.FAILED:
+      return NotificationTransactionStatus.FAILED;
+    case TransactionStatus.EXPIRED:
+      return NotificationTransactionStatus.EXPIRED;
+    case TransactionStatus.CANCELLED:
+      return NotificationTransactionStatus.CANCELLED;
+    default:
+      throw new Error(`Unsupported status: ${status}`);
```
xeno097, line 19: I think that it would be better to create an error using the `BaseInternalError` here instead. Example:
image

src/modules/notifications/infra/database/postgres/models/notifcation.model.ts
```diff
@@ -0,0 +1,22 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+@Entity({
+  // It must be the same as the UserDBModel as this is only a logical entity
+  name: 'user',
+})
+export class NotificationDBModel {
```
xeno097, line 7: I'd rename this to `UserNotificationInfoDBModel`. This applies also to the repository

src/modules/notifications/infra/database/postgres/models/notifcation.model.ts
```diff
@@ -0,0 +1,22 @@
...
+  @PrimaryColumn()
+  id: string;
+
+  @Column()
+  walletAddress: string;
```
xeno097, line 12: If the wallet address is not used in any notification I'd remove this from here

src/modules/notifications/infra/repositories/notifications.repository.ts
```diff
@@ -0,0 +1,43 @@
...
+import { InjectRepository } from '@nestjs/typeorm';
+import { EntityId } from 'src/core/domain/entity/entity-id';
+import { NotFoundError } from 'src/core/errors/not-found.error';
+import { Repository } from 'typeorm';
+import { NotificationEntity } from '../../core/domain/entity/notification.entity';
+import { DeviceToken } from '../../core/domain/value-objects/device-token.validator';
+import { INotificationRepository } from '../../core/infra/notification.repository';
+import { NotificationDBModel } from '../database/postgres/models/notifcation.model';
+
+const RESOURCE_NAME = 'NotificationRepository';
```
xeno097, line 11: I think that there was a bit of copy-pasting here with the name

src/modules/notifications/notification-message.interface.ts
```diff
@@ -50,11 +52,23 @@ const metamapFailedMessage: Readonly<NotificationMessage> = {
     'Asegúrate de que tus documentos y tu información coincidan y estén actualizados. Si el problema persiste, contáctanos.',
 };
 
-export const trasnferNotification: Readonly<NotificationMessage> = {
+export const P2PNotification: Readonly<NotificationMessage> = {
   title: 'TE ACABA DE LLEGAR UN DINERITO 🤑',
   message: `!Has recibo $0 USDC de xxx (o una cartera externa)`,
 };
```
xeno097, line 58: if this has been replaced by the output of the function below, i believe we can remove this

src/modules/notifications/core/services/notification.service.ts
```diff
@@ -0,0 +1,34 @@
...
+    return this.withdrawNotificationUseCase.execute(dto);
+  }
+
+  public async sendMetamapNotification(dto: MetamapNotificationUseCaseDto): Promise<void> {
+    return this.metamapNotificationUseCase.execute(dto);
+  }
+
+  public async sendP2PNotification(dto: P2PNotificationUseCaseDto): Promise<void> {
+    return this.p2pNotificationUseCase.execute(dto);
+  }
```
xeno097, line 33: Did this come back from the dead? if you need this only for calling the use cases from the public API class then you can simply call the use cases directly from the Public API. Remember you can't call a use case from a service class

src/modules/notifications/core/infra/notification.repository.ts
```diff
@@ -0,0 +1,6 @@
+import { EntityId } from 'src/core/domain/entity/entity-id';
+import { NotificationEntity } from '../domain/entity/notification.entity';
+
+export abstract class INotificationRepository {
+  abstract getById(id: EntityId): NotificationEntity | Promise<NotificationEntity>;
```
xeno097, line 5: I think that it would be better to call it `getByUserId` just to make it clearer by which id we are filtering

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -1,49 +1,72 @@
...
+        {
+          provide: INotificationRepository,
+          useValue: {
+            getById: jest.fn().mockImplementation((id: EntityId) =>
+              Promise.resolve({
+                id: id.toString(),
+                deviceToken: VALID_DEVICE_TOKEN,
+              })
+            ),
+          },
```
xeno097, line 39: I'd prefer that you create a function that returns a mock class just as you did for the `INotificationRepository` and mock the return values in the tests that require it instead of setting the mocked implementation as default

src/modules/notifications/core/use-cases/test/deposit-use-case.test.ts
```diff
@@ -70,33 +95,41 @@ describe(DepositNotificationUseCase.name, () => {
...
+      name: 'DELETED_USER',
+      lastName: 'DELETED_USER',
+      walleltAddress: '0xB087ab8dd1A135E1ef9b35aA79Ed9982f6d37B22',
+      deviceToken: DeviceToken.fromString(INVALID_DEVICE_TOKEN),
+    });
+
+    // Act & Assert
+    await expect(depositUseCase.execute(dto)).rejects.toThrow(InvalidDeviceTokenError);
+
+    // Verifica que getById fue llamado correctamente, lo cual es útil para confirmar que el flujo de error es por no encontrar el usuario
```
xeno097, line 132: Please translate this comment to English

src/modules/notifications/core/use-cases/p2p-notification.use-case.ts
```diff
@@ -0,0 +1,51 @@
...
+   * @property {string} tokenSymbol - token symbol
+   */
+  tokenSymbol: SupportedCryptoCurrency;
+}
+
+@Injectable()
+export class P2PNotificationUseCase implements IUseCase<P2PNotificationUseCaseDto, void> {
+  //  TODO: Include userService and walletService
+  constructor(
+    private readonly firebaseNotificationProvider: INotificationsProvider,
```
xeno097, line 32: Same as the other comment about the variable naming

xeno097 (changes requested):

